PR #23, opened 2017-12-12
Feature/cmts hfc prov connection

schmto:


modules/HfcReq/Http/Controllers/NetElementController.php
```diff
@@ -20,18 +20,24 @@ public function view_form_fields($model = null)
 		$parents 	 = $model->html_list(NetElement::get(['id','name']), 'name', $empty_field);
 		$kml_files   = $model->kml_files();
 
+		// parse which netelementtype we want to edit/create
+		// NOTE: this is for auto reload via HTML GET
+		$type = 0;
+		if (isset($_GET['netelementtype_id']))
+			$type = $_GET['netelementtype_id'];
+		elseif ($model->netelementtype)
+			$type = $model->netelementtype->get_base_type();
```
NinoRy: $model->netelementtype should be checked, no?

modules/HfcReq/Http/Controllers/NetElementController.php
```diff
@@ -20,18 +20,24 @@ public function view_form_fields($model = null)
...
 
-		if ($model->netelementtype && $model->netelementtype->get_base_type() == 3) // cluster
+		if ($type == 3) // cluster
 			$prov_device = $model->html_list(\Modules\ProvBase\Entities\Cmts::get(['id', 'hostname']), 'hostname', $empty_field);
 
-		if ($model->netelementtype &&
-		   ($model->netelementtype->get_base_type() == 4 || // amp
-			$model->netelementtype->get_base_type() == 5)) // node
+		if ($type == 4 || $type == 5) // amp || node
 			$prov_device = $model->html_list(\Modules\ProvBase\Entities\Modem::get(['id', 'id']), 'id', $empty_field);
```
NinoRy: get id twice works and is wished behaviour?

modules/HfcReq/Http/Controllers/NetElementController.php
schmto: i guess no, because it is always initialized correctly via line:
$model = $model ? : new NetElement;

schmto: or better to say: Patrick’s generic checking of relations should work correctly during line  

elseif ($model->netelementtype)

modules/HfcReq/Http/Controllers/NetElementController.php
schmto: yes

PR #33, opened 2018-01-09
Creation of voip accounts possible if envia contract exists.

xee8ai:


xee8ai:
Changes in this pull request are contained in the next to create pull request handling the same topic.

PR #37, opened 2018-01-30
Bugfix/re creation of rejected voipaccounts

xee8ai:


modules/ProvVoip/Entities/Phonenumber.php
```diff
@@ -439,7 +439,15 @@ public function envia_contract_created() {
 			return $this->contract_external_id;
 		}
 		else {
-			return false;
+			// take the most recent contract from modem
+			// TODO: on handling of multiple contracts per modem: return all IDs
+			$envia_contract = $this->mta->modem->enviacontracts->last();
```
NinoRy: Just for general knowledge: i think
$envia_contract = $this->mta->modem->enviacontracts()->latest()->first();
has the same functionality but improves performance as this not instanciates multiple enviacontracts inside the collection through which you iterate then. In this case it will not really matter as a contract probably will not have so many enviacontracts

PR #39, opened 2018-02-07
Enabled auth middleware for datatable routes.

xee8ai:
@cschra: Using auth middleware is essential. Please verify that my change does not affect datatables functionality (AFAICS it should not. Tests are running, too.).

### DIFF
--- a/app/extensions/core/BaseRoute.php
+++ b/app/extensions/core/BaseRoute.php
@@ -84,7 +84,7 @@ class BaseRoute {
 		\Route::get(strtolower($name).'/fulltextSearch', array('as' => $name.'.fulltextSearch', 'uses' => $controller.'@fulltextSearch', $options, 'middleware' => 'auth:view'));
 
 		// AJAX Index DataTable
-		\Route::get("$name/datatables", array('as' => $name.'.data', 'uses' => $controller.'@index_datatables_ajax', $options));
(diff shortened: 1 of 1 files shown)


cschra:
works, but to keep future api development without auth this is possible:
(App::environment('APP_DEBUG') ? 'middleware' : '') => (App::environment('APP_DEBUG') ? 'auth:view' : '')));

This turns off middleware when in development mode - just a suggestion

schmto:
I think disabling middleware/auth in DEBUG mode is not this cool. Why not instead using a separate ENV variable. Normally this variable is not set. This implies that the user who sets this variable knows what he is doing. Turning DEBUG mode on and losing middleware/auth is quit risky. In order many people use DEBUG mode, only to see error on WEB in a cool form. Nobody thinks about, that this will shutdown middleware.

cschra:
DEBUG is standard ENV Variable for development provided by Laravel.
Running an App in Debug Mode as a standarad is not recommended... I would rather recommend to make it a global config setting so (only) an admin could enable it - but it should be disabled as standard in production.

xee8ai:
I do not understand why auth middleware should be disabled in debug mode (except that it bloats the debug data a bit). Disabling it finally results in debugging other code than that running in production – which may cause problems.

cschra:
to be honest: we dont use datatables as real api and therefor auth-middleware is fine and your points are valid. If we want to do it right, an Authentification with JSON Webtoken is neccessary (in other words: a special API middleware)
But I already mentioned that enabling Middleware only in debug is not clever and we should not do that.
However we should not run in debug mode on client systems as standard... thats all I was saying

PR #44, opened 2018-02-12
Feature/lar 100 frontend improvements christian

cschra:
Reworked global config, to show validation errors again and changed Hyperlink-Color according to aggrement with @NinoRy


schmto:
thx, we will review and test it tomorrow! Then it will directly go in production. Did you test it well?

PR #55, opened 2018-02-15
Bugfix/prior update2 ole

NinoRy:


olebowle:
not checked in yet, still fixing bugs ;-)

schmto:
yes the question was assigned to commit be960ea – @NinoRy

NinoRy:
is tested and works

PR #64, opened 2018-02-23
Import: Discard voip tariffs of contracts without mta assigned

NinoRy:


NinoRy:
On old system voip tariffs of contracts without mta were not charged ... so we can/have to just discard these tariffs for our system

PR #91, opened 2018-04-04
Bugfix/endpoint and testing

xee8ai:
The Endpoint stuff crashed the unit testing (mainly the seeder) – fixed it.

I introduced new behavior: If fixed_ip is not checked then IP in database is set to null (EndpointObserver methods creating() and updating()). Otherwise this IP will silently be blocked from reuse.
@olebowle Are you with me on this change? Or is there a need to keep an IP address bundled with an endpoint?


olebowle:
looks fine to me

PR #104, opened 2018-04-11
Hid log_level form field in global config.

xee8ai:
Field is ATM not used to change log level (done via config file).


xee8ai:
Maybe there are more config options that have no influence on NMSPrime behavior – I suggest to hide them (until implemented), too.

PR #118, opened 2018-05-08
Bugfix/post php71

olebowle:


xee8ai:
I suggest two more fixes:
* add source command to cat /etc/cron.d/nmsprime (ATM cron jobs are using PHP 5.6 and not able to run)
* check if installing PHP 5.6 is still needed (e.g. INSTALL-REPO.sh, install-from-git.sh)

PR #124, opened 2018-05-22
Bugfix/small fixes nino

NinoRy:


modules/ProvMon/Http/Controllers/ProvMonController.php
```diff
@@ -5,7 +5,7 @@
 
 use View;
 use Acme\php\ArrayHelper;
-use Modules\ProvBase\Entities\{ Cmts, Configfile, Endpoint, IpPool, Modem, ProvBase, ProvVoip, Qos };
```
schmto: ProvVoIP is not a MUST module. So we should check instead if module is enabled

modules/ProvMon/Http/Controllers/ProvMonController.php
```diff
@@ -728,16 +728,18 @@ public function search_lease ()
...
+		$file = file_get_contents('/var/lib/dhcpd/dhcpd.leases');
+		// start each lease with a line that begins with "lease" and end with a line that begins with "{"
+		preg_match_all('/^lease(.*?)(^})/ms', $file, $section);
 
-			// fetch all lines matching hw mac
-			foreach (array_unique($section[0]) as $s)
+		// fetch all lines matching hw mac
+		foreach (array_unique($section[0]) as $s)
+		{
+			if (strpos($s, $search))
```
schmto: IMHO this makes no sense, because the if statement is still there. I guess a type mistake?

PR #126, opened 2018-05-23
Bugfix/post php71

olebowle:


olebowle:
This should be approved/merged prior to going to dev on mablx10.

PR #141, opened 2018-05-31
small fixes

NinoRy:


modules/Ticketsystem/Entities/Ticket.php
```diff
@@ -87,6 +91,12 @@ public function username()
 		return $this->user ? $this->user->first_name.' '.$this->user->last_name : '';
 	}
 
+	public function tickettype_names_query()
+	{
+		return "group_concat(tickettypes.names separator ', ') like ?";
+		// from ticket t, tickettype_ticket ttt, tickettype tt where t.id=ttt.ticket_id and ttt.tickettype_id=tt.id";
```
cschra: what does this comment do and does it need to stay here?
Is the filter necessary for search? names should be able to be searched via eager loading... or is only the "first" name accessible?

PR #165, opened 2018-07-05
Bugfix/small fixes ole

olebowle:


Install/module_after_upgrade.sh
```diff
@@ -8,6 +8,7 @@ cd "$dir"
 /opt/rh/rh-php71/root/usr/bin/php artisan module:migrate
 /opt/rh/rh-php71/root/usr/bin/php artisan queue:restart
 /opt/rh/rh-php71/root/usr/bin/php artisan nms:auth
```
cschra: nms:auth does not exist anymore. Please use the new Command  `auth:nms`

modules/HfcCustomer/Entities/ModemHelper.php
```diff
@@ -17,23 +17,23 @@ class ModemHelper extends \BaseModel {
 
 	public static function ms_num ( $s )
 	{
-		return Modem::whereRaw("$s AND us_pwr > 0")->count();
+		return Modem::where('netelement_id', '=' , $s)->where('us_pwr', '>', '0')->count();
```
cschra: for "=" you can skip the second where argument...
```php
return Modem::where('netelement_id', $s)->where('us_pwr', '>', '0')->count();
```

modules/HfcSnmp/Resources/views/NetElement/controlling.blade.php
```diff
@@ -22,6 +22,9 @@
 	@if ($reload)
 	<div class="row">
 		@DivOpen(10)
+			@if ($view_var->controlling_link)
+				{{ link_to($view_var->controlling_link, 'View...', ['class' => 'btn btn-primary pull-right']) }}
```
cschra: {!! !!}

PR #169, opened 2018-07-18
Feature/extend testing 2

xee8ai:


app/Http/Middleware/CheckApiUser.php
```diff
@@ -34,7 +34,7 @@ public function handle($request, Closure $next)
 	{
 		// only user api@localhost is allowed
 		if ($request->getUser() !== 'api@localhost')
-			return response()->json(['ret' => 'unauthorized']);
+			return response()->json(['ret' => 'unauthorized'], 401);
```
cschra: this middleware does not exist anymore with updates to 5.3.
So API Requests return the same as in the gui (401 if not logged in and 403 if logged in and no permission)
Pleasy have a look, if this still works like you expect it to...

PR #170, opened 2018-07-20
Laravel 5.3

cschra:
Update of Laravel and Introduction of new Permission System -> Bouncer
Also a lot of GUI Fixes and Code readability improvements
Deletion of Dead Code etc


modules/ProvBase/Database/Migrations/2018_05_09_000100_add_country_code_fields.php
```diff
@@ -3,9 +3,9 @@
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Database\Migrations\Migration;
 
-class AddCountryCodeFields {
+class AddCountryCodeFields extends Migration
```
cschra: @xee8ai this was created by you, but did not run through migration, so I fixed it. Did this Migration run on your Machine?

modules/ProvVoipEnvia/Http/routes.php
```diff
@@ -1,17 +1,40 @@
...
-	BaseRoute::get('/EnviaOrder/{EnviaOrder}/marksolved', array('as' => 'EnviaOrder.marksolved', 'uses' => 'Modules\ProvVoipEnvia\Http\Controllers\EnviaOrderController@mark_solved'));
-
 	BaseRoute::resource('EnviaContract', 'Modules\ProvVoipEnvia\Http\Controllers\EnviaContractController');
 	BaseRoute::resource('EnviaOrder', 'Modules\ProvVoipEnvia\Http\Controllers\EnviaOrderController');
 	BaseRoute::resource('EnviaOrderDocument', 'Modules\ProvVoipEnvia\Http\Controllers\EnviaOrderDocumentController');
+
+	Route::get('/provvoipenvia/cron/{job}', [
+		'as' => 'ProvVoipEnvia.cron',
+		'uses' => 'Modules\ProvVoipEnvia\Http\Controllers\ProvVoipEnviaController@cron',
+		// TODO: @xee8ai @Patrick: which auth middleware should be used?
```
cschra: @xee8ai  please have a look at this -> Might be good to move it to Commands and call it with Artisan!?

cschra:
These commits were generated by Laravel Shift and change indentation automatically ...  - i had no controll over these commits
the depricated folder should be removed in the later process...

cschra:
deprecated files are removed in e2c4e02

composer.json
```diff
@@ -5,24 +5,23 @@
...
+		"laravelcollective/bus": "^5.3.0",
 		"yajra/laravel-datatables-oracle": "6.*",
 		"markrogoyski/ipv4-subnet-calculator": "2.0.*",
 		"nwidart/laravel-modules": "^1.0",
-		"maatwebsite/excel": "^2.1",
 		"doctrine/dbal": "^2.7"
 	},
 	"require-dev": {
-		"phpunit/phpunit": "~5.7",
+		"phpunit/phpunit": "~4.0",
```
NinoRy: why do you go back to version 4.0 ? Should be a bug as 4.0 does not work with php7 if Patrick remembers this fact correctly

cschra: forgot to commit... have 5.7 on my machine - will add this to small fixes
don't know why or where I committed 4.0

app/Http/Controllers/WelcomeController.php
```diff
@@ -30,12 +30,29 @@ public function __construct()
...
+
+		if (env('APP_ENV') == 'testing') {
+			// $_SERVER['SERVER_PORT'] does not exist if running phpunit
+			$server_port = \Request::getPort();
+		}
+		else {
+			$server_port = $_SERVER['SERVER_PORT'];
+		}
+
+        if ($server_port == $admin_port)
```
NinoRy: This will often end in an exception as $server_port is not initialised when admin & ccc port is different! This needs to be fixed!

app/Http/Controllers/WelcomeController.php
```diff
@@ -30,12 +30,29 @@ public function __construct()
...
+            return redirect('admin/login');
+
+        if ($server_port == $ccc_port)
+            return redirect('customer');
 		if (\App::isLocal())
-			return view('welcome')->with(compact('head1', 'head2'));
```
NinoRy: pls dont commit these nonsense changes ;) where only subjective improvements where made

app/Http/Controllers/WelcomeController.php
```diff
@@ -30,12 +30,29 @@ public function __construct()
...
 	public function index()
 	{
-		$g = \GlobalConfig::first();
-		$head1 = $g->headline1;
-		$head2 = $g->headline2;
+        $admin_port = env('HTTPS_ADMIN_PORT', '443');
+        $ccc_port   = env('HTTPS_CCC_PORT', '443');
 
+        // if same port, show start page
+        if ($admin_port == $ccc_port)
```
NinoRy: Does this not often end in an exception as $server_port is not initialised when admin & ccc port is different ??

app/Http/Middleware/RedirectIfAuthenticated.php
```diff
@@ -1,44 +1,27 @@
...
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @param  string|null  $guard
+     * @return mixed
+     */
+    public function handle($request, Closure $next, $guard = null)
+    {
+        if (Auth::guard($guard)->check()) {
+            return redirect('admin');
```
NinoRy: Is it possible here to not always redirect to dashboard, but to the url from $request e.g. if someone wanted to /Modem/xxx/edit directly ?

app/Http/Controllers/WelcomeController.php
```diff
@@ -13,29 +14,26 @@ class WelcomeController extends Controller {
...
-		$head2 = $g->headline2;
+		$server_port = \Request::getPort();
+		$admin_port = env('HTTPS_ADMIN_PORT', '8080');
+		$ccc_port   = env('HTTPS_CCC_PORT', '443');
+
+		if ($server_port == $admin_port)
+				return redirect(route('adminLogin'));
+
+		if ($server_port == $ccc_port)
+				return redirect(route('customerLogin'));
```
cschra: @NinoRy  Final version of the Welcomecontroller. The ports are set via the Laravel Request Facade.
This logic did not belong into a middleware, because its only called once and needs a lot of time if applied on every request...

app/Http/Middleware/RedirectIfAuthenticated.php
cschra: This middleware is only used on the login page and has the purpose to redirect you to the Dashboard if you are allready have a Session.
Through the "can" middleware you are redirected to Login Page if you are not logged in and then redirected to your intended Page like "Modem/xy"  (due to an agreement with @schmto and @olebowle the new edit routes dont contain the edit word - just the id)

config/app.php
```diff
@@ -181,53 +183,57 @@
...
+        'DB' => Illuminate\Support\Facades\DB::class,
+        'Eloquent' => Illuminate\Database\Eloquent\Model::class,
+        'Event' => Illuminate\Support\Facades\Event::class,
+        'File' => Illuminate\Support\Facades\File::class,
+        'Gate' => Illuminate\Support\Facades\Gate::class,
+        'Hash' => Illuminate\Support\Facades\Hash::class,
+        'Lang' => Illuminate\Support\Facades\Lang::class,
+        'Log' => Illuminate\Support\Facades\Log::class,
+        'Mail' => Illuminate\Support\Facades\Mail::class,
+        'Notification' => Illuminate\Support\Facades\Notification::class,
```
NinoRy: why do you commit this? This just overloads the commit and makes it harder to review

modules/Ccc/Entities/CccUser.php
```diff
@@ -1,33 +1,39 @@
...
-	Authenticatable as AuthenticatableContract,
-	CanResetPassword as CanResetPasswordContract};
+namespace Modules\Ccc\Entities;
+
+use DB;
+use Illuminate\Notifications\Notifiable;
+use Silber\Bouncer\Database\HasRolesAndAbilities;
+use Illuminate\Auth\Authenticatable;
+use Illuminate\Auth\Passwords\CanResetPassword;
+use Illuminate\Foundation\Auth\Access\Authorizable;
```
NinoRy: are these 3 use statements not redundant?

config/app.php
cschra: Because this is the Laravel given style and in prep for 5.5 where automatic configuration is made when a new module installs, this style is much cleaner....

cschra: This commit was made when patrick wanted to checkouit to my branch and "to make it just work" i commited this, but forgort to cleanup the commit history, so  its here -> sorry again for the mess

app/Http/Controllers/BaseController.php
```diff
@@ -875,31 +870,56 @@ private static function _api_prepopulate_fields($obj, $ctrl) {
...
-		if (!$this->many_to_many)
-			return;
+		foreach ($this->many_to_many as $class => $field_name) {
 
-		foreach ($this->many_to_many as $field)
-		{
-			if (!isset($data[$field]))
-				$data[$field] = [];
+			if ( is_string($class) && \Bouncer::cannot('edit', $class ) )
+				abort(403, trans('You are not allowed to edit {$class}'));
```
NinoRy: this will never be translated. And $class will never be replaced by it's value ("" -> ''). But you must give $class as second parameter to the trans() function anyway!

app/extensions/html/FormBuilder.php
```diff
@@ -169,7 +169,7 @@ public function select($name, $list = array(), $selected = null, $options = arra
 
 		// Call the parent select method so that Laravel can handle
 		// the rest of the select set up.
-		if (isset($options['style']) && (strpos($options['style'], 'simple') !== false))
+		if (isset($options['style']) && (strpos($options['style'], 'simple') == true))
```
NinoRy: does this work? strpos never returns true and if it returns 0 it found the string at first position. I think this is wrong.

resources/views/Generic/relation.blade.php
```diff
@@ -68,8 +68,8 @@
 	<table class="table">
 		@foreach ($relation as $rel_elem)
 			<tr class="{{isset ($rel_elem->view_index_label()['bsclass']) ? $rel_elem->view_index_label()['bsclass'] : ''}}">
-				<td> {{ Form::checkbox('ids['.$rel_elem->id.']', 1, null, null, ['style' => 'simple']) }} </td>
-				<td> {{ $rel_elem->view_icon()}} {{ HTML::linkRoute($view.'.'.$method, is_array($rel_elem->view_index_label()) ? $rel_elem->view_index_label()['header'] : $rel_elem->view_index_label(), $rel_elem->id) }} </td>
+				<td width="20"> {{ Form::checkbox('ids['.$rel_elem->id.']', 1, null, null, ['style' => 'simple']) }} </td>
+				<td> {!! $rel_elem->view_icon() !!} {{ HTML::linkRoute($view.'.'.$method, is_array($rel_elem->view_index_label()) ? $rel_elem->view_index_label()['header'] : $rel_elem->view_index_label(), $rel_elem->id) }} </td>
```
NinoRy: why do you replace {{ by {!! sometimes and sometimes not?? on this line both variants are existent

app/extensions/html/FormBuilder.php
cschra: as this is no strict comparison (===) all positive and negative numbers convert to true and 0 converts to false. But I don't remember exactly why I changed this. - coud of cause be a bug or wrong logic...
While checking for strings I would rather use Str::contains(), because it checks for a string and deals with all the annoying edge cases...

resources/views/Generic/relation.blade.php
cschra: "{{  }}" fully escapes the content, while "{{!! !!}}" dont escapes the input, so when the content is html or has special characters we should use {!! !!} and if we have Text, we use {{ }}
This is a Blade security feature, if values get displayed that can be changed by the user, to prevent script attacks

cschra: normally {!! !!} should be avoided if possible, but due to our generic and how we compose the content, we often have to use them....

PR #196, opened 2018-08-22
Feature/add info on added voip item

xee8ai:


resources/views/Generic/relation.blade.php
```diff
@@ -41,7 +41,7 @@
 			{!! Form::close() !!}
 		@endif
 	@endcan
-	@can('delete', reset($relation))
+	@can('delete', $relation->get(0))
```
cschra: -> first() would be more elegant here in my opinion....

PR #233, opened 2018-08-31
Bugfix/small fixes christian

cschra:
Things discussed with @schmto and some small improvements...


modules/HfcBase/Http/Controllers/HfcBaseController.php
```diff
@@ -41,27 +41,23 @@ public function get_file($type, $filename)
...
-
-        return $a;
+        return $trees->filter(function ($tree) {
+            return $tree->kml_file != '';
+        })->map(function ($tree) {
+            return [
+                'file' => route('HfcBase.get_file', ['type' => 'kml_static', 'filename' => basename($tree->kml_file)]),
+                'descr' => $tree->kml_file,
+            ];
+        });
```
schmto: is this tested with multiple KML files showing on a topo card? In this case, all recursive KMLs should be displayed?

cschra: Logic wise and with a Demo array this Worked. Unfortunately there are no real kml files in the Demo Environment. So no, not production tested, but I am pretty confident with this...

PR #245, opened 2018-09-04
Now checking for freshly created modems to geocode.

xee8ai:


modules/ProvBase/Entities/Modem.php
```diff
@@ -1559,25 +1559,34 @@ public function updating($modem)
             return;
         }
 
+        // get changed values
         $diff = $modem->getDirty();
 
+        // if testing: do not try to geocode or position modems (faked data; slows down the process)
+        if (\App::runningUnitTests()) {
```
cschra: better check for App environment "testing" ?

PR #251, opened 2018-09-06
Feature/offical help integration 2

schmto:


cschra:
please commit the style changes... you can get a patch file directly from styleCI.

modules/Dashboard/Http/Controllers/DashboardController.php
```diff
@@ -541,23 +541,113 @@ private static function _get_perfdata_class($cur, $warn, $crit)
...
      * process is and addvice the next steps the user should do..
      *
      * This function could also be used to inform the user of new updates (etc)
-     *
-     * @TODO: This should be somehow cashed and must not always be run when loading dashboard!
      */
     public function news()
+    {
+        // check for insecure install
+        $sec = $this->news_is_insecure_install();
```
cschra:
```php
$sec = $this->news_is_insecure_install() ? return $this->news_is_insecure_install() : false;
````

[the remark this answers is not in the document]
Comment (cschra) [reply modules/Dashboard/Http/Controllers/DashboardController.php]: I would think about changing the logic here... Doing negative checking -> return true if it is insecure and false if its not.
Makes more sense to me...

File: modules/Dashboard/Http/Controllers/DashboardController.php [hunk 2]
```diff
@@ -541,23 +541,113 @@ private static function _get_perfdata_class($cur, $warn, $crit)
...
+        if ($sec)
+            return $sec;
+
+        // Install add sequence check
+        if (\Module::collections()->has('ProvBase'))
+            if (\Modules\ProvBase\Entities\Modem::count() == 0)
+                return $this->news_install_add_sequence_check();
+
+        // Check for official news from nmsprime.com
+        $news = $this->news_load_official_site();
```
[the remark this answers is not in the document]
cschra: same as above

modules/Dashboard/Http/Controllers/DashboardController.php
```diff
@@ -605,20 +695,30 @@ public function news()
...
-        if (env('ROOT_DB_PASSWORD') == '') {
-            try {
-                \DB::connection('mysql-root')->getPdo();
-                if (\DB::connection()->getDatabaseName()) {
-                    return ['youtube' => 'https://www.youtube.com/embed/dZWjeL-LmG8',
-                    'text' => '<li>Danger! Run: mysql_secure_installation in bash as root!', ];
-                }
-            } catch (\Exception $e) {
+        // add costcenter
+        if (\Module::collections()->has('BillingBase')) {
```
cschra:
```php
 if (\Module::collections()->has('BillingBase') && \Modules\BillingBase\Entities\CostCenter::count() == 0) {
```

modules/Dashboard/Http/Controllers/DashboardController.php
```diff
@@ -605,20 +695,30 @@ public function news()
...
+                return ['youtube' => null,
+                    'text' => '<li>Next: '.\HTML::linkRoute('CostCenter.create', 'Create a first Cost Center'), ];
             }
         }
 
-        // links need to be in embedded style, like:
-        //return ['youtube' => 'https://www.youtube.com/embed/9mydbfHDDP4',
-        //		'text' => "You should do: <a href=https://lifeisgood.com>BlaBlaBla</a>"];
+        // add Contract
+        if (\Module::collections()->has('ProvBase')) {
```
cschra: ...same as above

modules/Dashboard/Http/Controllers/DashboardController.php
```diff
@@ -605,20 +695,30 @@ public function news()
...
+            if (\Modules\ProvBase\Entities\Contract::count() == 0) {
+                return ['youtube' => 'https://www.youtube.com/embed/t-PFsy42cI0?start=0&',
+                    'text' => '<li>Congratulations: now you can create a first '.\HTML::linkRoute('Contract.create', 'Contract'), ];
+            }
+        }
+
+        // add Modem
+        if (\Module::collections()->has('ProvBase')) {
```
cschra: same as above

modules/Dashboard/Http/Controllers/DashboardController.php
```diff
@@ -541,23 +541,113 @@ private static function _get_perfdata_class($cur, $warn, $crit)
...
     {
         // change default psw's
         if (\Hash::check('toor', \Auth::user()->password)) {
             return ['youtube' => 'https://www.youtube.com/embed/TVjJ7T8NZKw',
                     'text' => '<li>Next: Change default Password! '.\HTML::linkRoute('User.profile', 'Global Config', \Auth::user()->id), ];
         }
 
+        // check for insecure MySQL root password
+        // This requires to run: mysql_secure_installation
+        if (env('ROOT_DB_PASSWORD') == '') {
```
cschra:
```php
if (env('ROOT_DB_PASSWORD') != '') {
    return;
}
try {
    if ( \DB::connection('mysql-root')->getPdo()&& \DB::connection()->getDatabaseName()) {
        return ['youtube' => 'https://www.youtube.com/embed/dZWjeL-LmG8',
                    'text' => '<li>Danger! Run: mysql_secure_installation in bash as root!', ];
    } catch (\Exception $e) {
// ??? throw exception or do some output here
}

```

modules/Dashboard/Http/Controllers/DashboardController.php
```diff
@@ -541,23 +541,113 @@ private static function _get_perfdata_class($cur, $warn, $crit)
...
+     */
+    private function news_load_official_site()
+    {
+        $json = json_encode([
+            'youtube' => "",
+            'text' => ""
+        ]);
+
+        $ch = curl_init();
+        curl_setopt($ch, CURLOPT_URL, 'https://repo.roetzer-engineering.com/rpm/nmsprime-news/index.json');
```
cschra: sure you want a static URL here? -> better set this via config or env

modules/Dashboard/Http/Controllers/DashboardController.php
cschra: if you dont catch the error, there is no reason to use try- catch statement... either you throw an error or do logging or something else. But if nothing is done during an exception, the try-catch could be removed....

modules/Dashboard/Http/Controllers/DashboardController.php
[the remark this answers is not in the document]
cschra: @olebowle pointed out that returning in a ternary might also be not optimal or good style... so maybe change it to 
```php
if ($insecure = this->news_is_insecure_install()) {
    return $insecure;
}
```

modules/Dashboard/Http/Controllers/DashboardController.php
schmto: IMHO: a try without catch does not break the code if a problem happens inside. If code breaks without a try, we will see an error here. This was my goal here

schmto:
@cschra : please review again.

modules/Dashboard/Http/Controllers/DashboardController.php
schmto: AND: i use the try inside the if-statement to let other checks possible run afterwards. Your proposal will break this workflow and skip other tests

cschra: The if statement is inverted to better indent the code and to clear the code.... because this is essentially what you do. If the root_db_password is not empty, you return from this method.
```php
if (env('ROOT_DB_PASSWORD') != '') {
    return;
}
```
And try with empty catch is an anti-pattern (see [here ](https://stackoverflow.com/questions/4820211/catch-an-exception-without-doing-anything-in-the-catch), [here](https://softwareengineering.stackexchange.com/questions/16807/is-it-ever-ok-to-have-an-empty-catch-statement/16822) or [here](https://stackoverflow.com/questions/17838143/try-with-empty-catches)), you should at least log the exception. Because if it fails, its is for a reason
I dont know, if this is 100% correct, but would make more sense to me:
```php
try {
    \DB::connection('mysql-root')->getPdo();
    \DB::connection()->getDatabaseName()
    } catch (Exception $e) {
        return [
            'youtube' => 'https://www.youtube.com/embed/dZWjeL-LmG8',
            'text' => '<li>$e->message - Run: mysql_secure_installation in bash as root!', 
        ];
}
```

PR #271, opened 2018-09-14
Feature/cmts diagrams

olebowle:


schmto:
i don't get the point: how is the cluster specific association done? or is it straight forward: showing "only" all cacti diagrams?

olebowle:
Please have a look at line [921](https://github.com/nmsprime/nmsprime/pull/271/files#diff-d1d836e801b339c474d91d3cf0299b28R921) and line [934](https://github.com/nmsprime/nmsprime/pull/271/files#diff-d1d836e801b339c474d91d3cf0299b28R934). First I get all the indices of the netelement in question and filter for those in the cacti db.

schmto:
ah, cool. which means: it's a combination of netelement linking AND generic SNMP GUI module(?). Correct? Then i got the point.

schmto:
more than perfect approach :+1:

PR #510, opened 2018-10-11
Changes to suggest mac address of mta. 

kokuta:


cschra:
please visit styleCI and apply the changes to your branch.
thanks for contributing to nmsprime

schmto:
@kokuta IMHO: it is not working with leading zeros
Examples:
Modem MAC –> MTA MAC
0a:11:02:33:41:2a –> A1:10:23:34:12:B (should be 0a:11:02:33:41:2b)
00:11:22:33:44:55 –> 11:22:33:44:56 (should be 00:11:22:33:44:56)

PR #535, opened 2018-11-08
Feature/settlementrun adaptions nino

NinoRy:


NinoRy:
We need to update some tariffs of ERZNETs production environment after these changes! (add minimum maturity, and maturity)
Use this SQL Query to see what contracts:
SELECT c.number, c.lastname, c.contract_start, i.valid_from, i.valid_to, p.name, p.type FROM contract c, item i, product p where c.id=i.contract_id and 
				p.id=i.product_id and p.type in ('Internet') and i.deleted_at is null and c.id in (select contract_id from item, product where product.id=item.product_id and
				product.type='Internet' and item.valid_from>CURRENT_DATE) and p.name like '%Jahre%' order by c.number;

PR #553, opened 2018-11-19
bugfix/small-fixes-ole

olebowle:


modules/ProvMon/Resources/views/cpe_analysis.blade.php
```diff
@@ -51,8 +51,11 @@
 
 @section('content_configfile')
 	@if (isset($configfile))
-		<font color="green"><b>{{$type}} Configfile</b></font><br>
-		@foreach ($configfile as $line)
+		<font color="green"><b>{{$type}} Configfile ({{$configfile['mtime']}})</b></font><br>
```
NinoRy: why is there a configfile in cpe_analysis blade? i think we can remove this part!?

PR #559, opened 2018-11-27
Modul CCC, small improvments

gaertnermarkus:
HTML-Typo, and change page title for customer


NinoRy:
Hey. Before we can merge this, the english translation file has to be synchronized with the german one. So english translation has to be there as well. And i would prefer the translations in resources/lang/{en,de}/messages.php as dt_header.php is intended for DataTable Header translations.

schmto:
FYI: @gaertnermarkus : @NinoRy comment is totally correct. I missed this. This step is required for auto-merge of https://crowdin.com/project/nmsprime language tool. Otherwise it will break changes if english translation is missing.

NinoRy:
> FYI: @gaertnermarkus : @NinoRy comment is totally correct. I missed this. This step is required for auto-merge of https://crowdin.com/project/nmsprime language tool. Otherwise it will break changes if english translation is missing.

Nope. I guess this wont work, as the string inside the trans function must begin with the language file name -> in this case 'messages.ccc.page_html_header'. Also AFAIK dots are used to seperate arrays in the files. Why not just use 'messages.ccc' as the translation already exists?

NinoRy:
> @NinoRy i have seen that other translation (ccc). And it's the easiest way to use this. But how can i assume that this is's used in not other way? So i decided to create a new one, with o more detailed title.

That's a valid objection. It's better to create a new one here.

Now we can merge it.
Thank you for your effort!

gaertnermarkus:
fine. your are welcome

PR #598, opened 2019-01-29
BugFix: Cache Netelements correctly

cschra:


modules/HfcReq/Entities/NetElement.php
```diff
@@ -519,7 +518,7 @@ public function updating($netelement)
             return;
         }
 
-        if ($netelement['original']['parent_id'] != $netelement['attributes']['parent_id']) {
+        if ($netelement->isDirty('parent_id') || $netelement->isDirty('name')) {
```
NinoRy: this loops twice through all columns ... store in variable first or just leave it as is for that single case

modules/HfcReq/Entities/NetElement.php
```diff
@@ -518,7 +518,7 @@ public function updating($netelement)
             return;
         }
 
-        if ($netelement->isDirty('parent_id') || $netelement->isDirty('name')) {
+        if ($netelement->isDirty('parent_id', 'name')) {
```
NinoRy: multiple use of isDirty again - pls store in variable

modules/HfcReq/Entities/NetElement.php
```diff
@@ -529,14 +529,14 @@ public function updating($netelement)
...
+        if ($netelement->isDirty('netelementtype_id')) {
             $new_params = $netelement->netelementtype->parameters;
 
-            foreach ($netelement->indices as $indices) {
-                // assign each indices of parameter to new parameter with same oid
-                if ($new_params->contains('oid_id', $indices->parameter->oid->id)) {
-                    $indices->parameter_id = $new_params->where('oid_id', $indices->parameter->oid->id)->first()->id;
-                    $indices->save();
+            foreach ($netelement->indices as $indice) {
+                // assign each indice of parameter to new parameter with same oid
```
NinoRy: why do you change this ? each entry contains really multiple indices

modules/HfcReq/Entities/NetElement.php
cschra: as discussed, this is fixed later on

PR #599, opened 2019-02-05
Enhanced contract form validation and some translations.

xee8ai:
firstname, lastname and birthday not longer required if salutaltion is “Firma” or “Behörde”, company now required on those.

@cschra Please check de and en validation translations. I sorted the attributes (before adding my translations) – does this overwrite already translated crowdin stuff (or is it smart enough)?


NinoRy:
@xee8ai Is the validation not just doable by laravels 'required_if' validation rule??

cschra:
> @cschra Please check de and en validation translations. I sorted the attributes (before adding my translations) – does this overwrite already translated crowdin stuff (or is it smart enough)?

we will see if this merges correct, but from my point of view it looks okay and it should(!) work.
Also I highly recommend using laravels native validation rule as mentioned by @NinoRy 😉

PR #600, opened 2019-02-06
Bugfix/restructure daily conversion nino

NinoRy:


app/helpers.php
```diff
@@ -238,3 +238,30 @@ function checkLocale($locale = null) : string
...
+ * where the date is larger/later or equal then the specified date
+ * NOTE: For end dates an empty column is later - it's cumbersome to always write these 5 lines of code
+ *
+ * @param string    db column name - must be table.column in joined statements
+ * @param string    date string like '2019-02-06' - default is today
+ * @return function db query to use in (chained) where clause
+ *
+ * @author Nino Ryschawy
+ */
+function whereLaterOrEqualThanDate($column, $date = '')
```
cschra: would change the name to `laterOrEqualThan` - reads more fluent - but this is purely personal opinion

modules/ProvBase/Console/contractCommand.php
```diff
@@ -40,11 +40,42 @@ public function __construct()
...
+            if (\Module::collections()->has('BillingBase')) {
+                // attention: do not check if valid_to or valid_from is fixed – we need them both
+                //      (change item dates, trigger online state of modems)
+                $cs = Contract::where(whereLaterOrEqualThanDate('contract_end', $min_date))
+                    ->where('contract_start', '<=', $today)
+                    ->join('item', 'contract.id', '=', 'item.contract_id')
+                    ->join('product', 'product.id', '=', 'item.product_id')
+                    ->whereIn('product.type', ['Internet', 'Voip'])
+                    ->whereBetween('item.valid_from', [$min_date, $max_date])
+                    ->orWhereBetween('item.valid_to', [$min_date, $max_date])
```
NinoRy: Have you checked that the orWhereBetween doesnt break the scope of the single where clauses and gets items with product type 'Other', ... as well? This can happen because laravel could concatenate the where clauses like so: where product.type=... and item.valid_from=... or item.valid_to=... what means this: (product.type=... and item.valid_from=...) or item.valid_to=...

modules/ProvBase/Console/contractCommand.php
```diff
@@ -74,7 +105,7 @@ public function fire()
     protected function getArguments()
     {
         return [
-            ['date', InputArgument::REQUIRED, 'daily/monthly'],
+            ['date', InputArgument::REQUIRED, 'daily/daily_all_contracts/monthly'],
```
NinoRy: i would name the key just 'daily_all' ... as the name of the contractCommand already implies that it's related to contract and it's shorter and easier to write on command line

modules/ProvBase/Console/contractCommand.php
```diff
@@ -40,11 +40,42 @@ public function __construct()
...
-        if (! ($this->argument('date') == 'daily' || $this->argument('date') == 'monthly')) {
+        // we don't check contracts that ended before defined here (in days)
+        $days_around_start_and_enddates = 7;
+
+        $min_date = date('Y-m-d', strtotime("-$days_around_start_and_enddates days"));
+        $max_date = date('Y-m-d', strtotime("+$days_around_start_and_enddates days"));
+        $today = date('Y-m-d');
+
+        if (! in_array($this->argument('date'), ['daily', 'daily_all_contracts', 'monthly'])) {
+            echo "Wrong/missing argument\n";
```
NinoRy: Laravel added extra functions for this cases the colorizes the output also: i would prefer to use: $this->error()
Just have a look.

NinoRy:
@xee8ai Thx for digging that deep into it and helping to improve that important piece of code. I think we should review that with highest priority.

modules/ProvBase/Console/contractCommand.php
xee8ai: You're right – fixed.

modules/ProvBase/Console/contractCommand.php
xee8ai: No problem for me – changed :-)

PR #601, opened 2019-02-06
Added generic method to add usage to select values.

xee8ai:
This ATM is used to show how many modems/mta uses a configfile.


NinoRy:
I think it would be good to place a help message next to the select field so that everyone knows what the number behind the name of the configfile means

PR #603, opened 2019-02-11
Alternative configfile count

cschra:


cschra:
This should be merged into #601 
it is a more database agnostic and Laravel way of doing this task - merge it if you like the solution

cschra:
@xee8ai 
I like this solution personally very much, because it feels cleaner to me.
However, it comes with a speed cost, due to using eloquent. 

I compensated the lost speed by caching the result of the query. This could also be done with your approach and would result in a similar speed boost.

Here are the Timings of the different requests.
```
Your Query:   0.0058989524841309 s
My approach:  0.099140167236328 s
My approach with using Cache: 0.00053095817565918 s
```
I talked already with @NinoRy  - maybe the query could be optimized to reduce the loading time.
No hard feelings if you don't like to use my approach.

xee8ai:
Hi @cschra 
I talked to @NinoRy – we decided to use “my” version. The main reason is maintainability – one has do clear the caches in ModelUsingCountObserver::created|updated|deleted and in ModelToBeCountObserver, too.
Building the cache seems to be expensive, too – and has to be done regularly.

But thanks for your suggestion – thats just a nice piece of code :-)

cschra:
Okay, then... If you like, you maybe can recycle some of the code...

Just as a thought:
if someone put some more work into this, he/she could extract this into a trait (i.e.  "HasHtmlLists")  and let the trait handle the whole functionality (even Observables) - advantage is, that every html listing related code is in one place, BaseModel becomes cleaner and only the models that use html lists could use the trait...
- if you are open to the idea itself, we could keep this branch. I would like to work on this later this year or make it open for community contributions.
I think there is also potential to improve on the query - or manipulate the result  of your ("fast") query. However I would strive for a more database agnostic approach.

cschra:
keep the branch for later use but close this PR as the other version should be used

PR #608, opened 2019-02-13
Bugfix in creating endpoint with dynamic IP.

xee8ai:
#


olebowle:
This has already been fixed in a022b43. But thanks anyway!

PR #616, opened 2019-02-26
bugfix/small-fixes-nino

NinoRy:


modules/BillingBase/Console/SettlementRunCommand.php
```diff
@@ -817,20 +821,29 @@ private static function _get_customer_nrs()
...
 
-        return \DB::table('phonenumber as p')
-            ->join('mta', 'p.mta_id', '=', 'mta.id')
-            ->join('modem', 'modem.id', '=', 'mta.modem_id')
-            ->join('contract as c', 'c.id', '=', 'modem.contract_id')
-            ->where(function ($query) use ($registrar) {
+        if ($withEmptyRegistrar) {
+            $whereCondition = function ($query) use ($registrar) {
                 $query
                 ->where('sipdomain', 'like', "%$registrar%")
```
xee8ai: Wouldn't it be safer (and maybe faster) to match "%$registrar" instead

PR #618, opened 2019-03-05
Small table change

gaertnermarkus:
set width for coloums, 
set fix sequence for cdr and incoive,
add collapse on year

#

### DIFF
--- a/modules/Ccc/Resources/views/index.blade.php
+++ b/modules/Ccc/Resources/views/index.blade.php
@@ -22,7 +22,7 @@
 					</div>
 					<div class="panel-body fader d-flex flex-column" style="overflow-y:auto;@if($loop->first)@else display: none;@endif; height:100%">
 						<table class="table table-bordered">
-							@foreach($years as $month => $months)
+							@foreach($years as $months)
 								<tr>
 									<td class="{{$months['CDR']['bsclass']}}" align="center" style="width: 50%">
 										@if($months['CDR']['link'] != '')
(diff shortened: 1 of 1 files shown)


modules/Ccc/Resources/views/index.blade.php
```diff
@@ -22,23 +24,20 @@ class="btn btn-xs btn-icon btn-circle btn-warning d-flex"
 						<table class="table table-bordered">
 							@foreach($years as $month => $months)
 								<tr>
-									@foreach($months as $type => $invoice)
```
NinoRy: Does it really work without the foreach over the months?

gaertnermarkus: There is one foreach for the years, an the second forach(this) for the months.
Yes, it works. The key is not nessary. 

This is the structure:
Array
(
    [2019] => Array
        (         
            [1] => Array
                (
                    [CDR] => Array
                        (
                            [link] => Illuminate\Support\HtmlString Object
                                (
                                    [html:protected] => <a href="https://192.168.33.10/customer/home/download/95">01/2019-Evn</a>
                                )
                            [bsclass] => info
                        )

                    [INVOICE] => Array
                        (
                            [link] => Illuminate\Support\HtmlString Object
                                (
                                    [html:protected] => <a href="https://192.168.33.10/customer/home/download/94">01/2019</a>
                                )
                            [bsclass] => info
                        )
                )
        )
)

PR #622, opened 2019-03-11
Bugfix/small fixes nino

NinoRy:
#


NinoRy:
@schmto @cschra What do you think about commit cb82b71 ?? @olebowle and me think that 'Zugewiesene' is not really necessary or even somehow a bit 'stupid' (e.g. Zugewiesene EnviaTelAPI)

cschra:
I think Assigned is also sometimes misleading. - But to be honest, I wouldnt remove the words, but rather exchange it with something more meaningful.
How does "Related" ("Verknüpfte(s)"/"Zugehörige"?) sound to you?

NinoRy:
Note: I moved the Commit to another branch (see https://github.com/nmsprime/nmsprime/pull/623 
f4ca850) as it is no (critical) bugfix what this branch is intended for. I just added the next comment there!

PR #623, opened 2019-03-11
Feature/small fixes nino

NinoRy:
#

### DIFF
--- a/modules/BillingBase/Entities/SepaMandate.php
+++ b/modules/BillingBase/Entities/SepaMandate.php
@@ -62,7 +62,7 @@ class SepaMandate extends \BaseModel
     {
         // Dont check when table header of index page is assembled
         if (! $this->id) {
-            return;
+            return '';
         }
 
         if ($this->check_validity('now')) {
--- a/modules/BillingBase/Entities/SepaMandate.php
+++ b/modules/BillingBase/Entities/SepaMandate.php
@@ -49,18 +49,31 @@ class SepaMandate extends \BaseModel
                 'index_header' => [$this->table.'.sepa_holder', $this->table.'.sepa_valid_from', $this->table.'.sepa_valid_to', $this->table.'.reference'],
                 'bsclass' => $bsclass,
                 'order_by' => ['0' => 'asc'],
-                'header' =>  "$this->reference - $this->sepa_iban", ];
+                'header' =>  "$this->reference - $this->sepa_iban - $this->sepa_valid_from $valid_to",
+            ];
     }
 
+    /**
+     * Get bootstrap class for colorization of model in GUI
+     *
+     * @return string
+     */
     public function get_bsclass()
     {
-        $bsclass = 'success';
-
-        if (isset($this->created_at) && ($this->get_start_time() > strtotime(date('Y-m-d'))) && ! $this->check_validity('Now')) {
-            $bsclass = 'danger';
+        // Dont check when table header of index page is assembled
+        if (! $this->id) {
+            return;
         }
 
-        return $bsclass;
+        if ($this->check_validity('now')) {
+            return 'success';
+        }
+
+        if ($this->get_start_time() > strtotime(date('Y-m-d'))) {
+            return 'info';
+        }
+
+        return 'active';
     }
 
     public function view_belongs_to()
--- a/modules/ProvBase/Entities/Contract.php
+++ b/modules/ProvBase/Entities/Contract.php
@@ -111,6 +111,7 @@ class Contract extends \BaseModel
     public function view_has_many()
     {
         if (\Module::collections()->has('BillingBase')) {
+            // view has many version 1
             $ret['Edit']['Modem'] = $this->modems;
             $ret['Edit']['Item'] = $this->items;
             $ret['Edit']['SepaMandate'] = $this->sepamandates;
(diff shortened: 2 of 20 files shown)


NinoRy:
@schmto @cschra @xee8ai 

Note: This comment refers to the comments from branch https://github.com/nmsprime/nmsprime/pull/622. I just move the commits here because of the reason explained in last comment there.

To the topic: 
What else should the user think when he sees 3 modems in the panel 'Modems' on the contract page? That they are related to to another contract? I think it's just superfluous and nobody would expect modems from another contract there. But that's just my opinion.

cschra:
I just clicked through the demo system to see if it really mattered and found the "Assigned/Zugewiesene" not necessary. It is indeed self expainatory.

resources/views/bootstrap/panel.blade.php
```diff
@@ -4,13 +4,15 @@
...
+
+	$display = isset($options['display']) ? 'display: '.$options['display'] : '';
 ?>
 
 {{-- begin col-dyn --}}
 <div class="col-md-{{$md}} ui-sortable">
 	<div class="panel panel-inverse card-2 d-flex flex-column" data-sortable-id="table-index-{{ isset($i) ? $i : '1'}}">
 		@include ('bootstrap.panel-header', ['view_header' => $view_header])
-		<div class="panel-body fader d-flex flex-column" style="overflow-y:{{ $overflow_y }}; {{ $style }}">
+		<div class="panel-body fader d-flex flex-column" style="overflow-y:{{ $overflow_y }}; {{ $style }}; {{ $display }}">
```
cschra: If you touch this part of the code, i would recommend unifying it.
Also as a personal preference I would inline the php stuff
```php
    <div class="panel-body fader d-flex flex-column" style="{{ $overflowY }}; {{ $style }}; {{ $display }}">
// or 
    <div class="panel-body fader d-flex flex-column" 
             style="overflow-y:{{ $overflowY }}; height: {{ $style }};display: {{ $display }}">
// or (my prefered option (with removal of the php part above) - as it is only used once
    <div class="panel-body fader d-flex flex-column" 
             style="overflow-y: {{ $overflow ?? 'auto' }}; 
                     height: {{ isset($height) ? (is_numeric($height) ? $height.'%' : 'auto') : '100%' }};  // not pretty, but it does the trick
                     display: {{ $options['display'] ?? 'inherit' }};"> // unset or initial would also be an option
```

modules/ProvMon/Resources/views/analyses.blade.php
```diff
@@ -1,17 +1,18 @@
...
 @section('content_dash')
+	<div class="btn pull-right">
+		@include('Generic.documentation', ['documentation' => $modem->help])
+	</div>
+
 	@if ($dash)
-		@foreach ($dash as $info)
-			<div class="note note-{{$info['bsclass']}} fade in m-b-15"> {{ $info['text'] }} </div>
+		@foreach ($dash as $key => $info)
+			<div class="alert alert-{{$info['bsclass']}} fade show" style="width: 83%"> {{ $info['text'] }} </div>
```
cschra: where do the 83% come from? why not use a css class like "col-md-9"?

modules/ProvMon/Http/Controllers/ProvMonController.php
```diff
@@ -131,11 +131,19 @@ public function analyses($id)
 
         // Log dhcp (discover, ...), tftp (configfile or firmware)
         $search = $ip ? "$mac|$modem->hostname|$ip " : "$mac|$modem->hostname";
+
+        // NOTE: This function takes a long time if syslog file is large - 0.4 to 0.6 sec
```
cschra: use caching for this?

modules/ProvBase/Entities/Contract.php
```diff
@@ -119,14 +120,23 @@ public function view_has_many()
...
 
         if (\Module::collections()->has('BillingBase')) {
+            // view has many version 2
             $ret['Billing']['Item']['class'] = 'Item';
             $ret['Billing']['Item']['relation'] = $this->items;
             $ret['Billing']['SepaMandate']['class'] = 'SepaMandate';
             $ret['Billing']['SepaMandate']['relation'] = $this->sepamandates;
+
+            // Show invoices in 2 panels - 2nd panel with old invoices collapsed and in 2 columns
+            $p1 = $this->invoices()->orderBy('id', 'desc')->take(15)->get();
```
cschra: why 15? wouldn't it be good to have a protected variable/ config / environment value set these limits?

resources/views/bootstrap/panel.blade.php
NinoRy: i dont want to rewrite the whole code. I just wanted to add the display attribute. Rewriting the whole code is another task and would make unclear what happened in the commit. I would have to make another commit at least. We could definitely better structure this blade like the most. But then we should do this in a consistent way with all the blades. That`s more like a rewrite of the existing blade structure what we of course should do somewhen

modules/ProvMon/Resources/views/analyses.blade.php
NinoRy: (y) didn't know it

modules/ProvMon/Http/Controllers/ProvMonController.php
NinoRy: caching doesnt really make sense here ... it's just a comment in a first step to be able to have a quick overview what takes a lot time ... for future performance works

modules/ProvBase/Entities/Contract.php
NinoRy: we could discuss this ... this is just quick and dirty ;) ... let us wait for feedback of the accounting staff

modules/BillingBase/Entities/SepaMandate.php
```diff
@@ -49,18 +49,31 @@ public function view_index_label()
...
+     * Get bootstrap class for colorization of model in GUI
+     *
+     * @return string
+     */
     public function get_bsclass()
     {
-        $bsclass = 'success';
+        // Dont check when table header of index page is assembled
+        if (! $this->id) {
+            return;
```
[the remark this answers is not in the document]
NinoRy: done

PR #637, opened 2019-03-27
Bugfix/small fixes nino

NinoRy:
#


NinoRy:
Let's just test if
 tac <file> | grep <string> -m 1
is faster than
 grep <string> <file> | tail -1
bevor we merge this. Or we think about another option like logging via firewalld, etc

PR #643, opened 2019-03-31
Feature/modules dashboards

kokuta:
#


public/components/nmsprime.css
```diff
@@ -461,3 +461,28 @@ table .index_check {
     height: 20px;
     background: #d9e0e7;
 }
+
```
NinoRy: This still doesn't seem to work. Do i miss sth? I executed php artisan view:clear and cache:clear, logged out and in again and cleared the browser cache. But it still loads the dashboard when i click the carot instead just expanding the submenu.

app/helpers.php
```diff
@@ -287,3 +287,29 @@ function clearFailedJobs($command = 'all')
...
+
+    switch (\App::getLocale()) {
+        case 'de':
+            return date('d.m.Y', $date);
+
+        case 'es':
+            return date('d/m/Y', $date);
+
+        default:
+            return date('d-m-Y', $date);
```
NinoRy: (1) Default must be 'Y-m-d'
(2) Does it always take the default here? I set my language to german and it still has the default format

modules/Dashboard/Http/Controllers/DashboardController.php
```diff
@@ -16,7 +16,7 @@ public function index()
     {
         $title = 'Dashboard';
 
-        $logs = GuiLog::where('username', '!=', 'cronjob')->orderBy('updated_at', 'desc')->limit(20)->get();
+        $logs = GuiLog::where([['username', '!=', 'cronjob'], ['model', '!=', 'User']])->orderBy('updated_at', 'desc')->limit(20)->get();
```
NinoRy: Is this supposed to work? Never saw this syntax. And i still have the 'Updated user' entries in the global dashboard. So in my case it doesn't seem to work.

kokuta: Since Laravel 5.2 there is possible to pass an array of conditions to where clause.
You can check the documentation of Laravel 5.3(our version) https://laravel.com/docs/5.3/queries#where-clauses
The fact that you still get the 'updated User' entry on the dashboard means that there might be problems with synchronizing the changes. 
Can you please double check synchronizing changes because none of the reported issues above appear at my local machine.

NinoRy: Sorry. My fault. Looks great.

NinoRy:
1 more question: Could you test the navigation sidebar on a mobile device? Just in case the carot is maybe too small to "hit" or sth like that. And would it be possible to enlarge the clickable area of the caret? (Probably with another div around it?)

resources/views/bootstrap/sidebar.blade.php
```diff
@@ -33,15 +33,15 @@
                     <i class="fa fa-fw {{ $typearray['icon'] }}"></i>
                     <span>{{$typearray['translated_name'] ?? $module_name}}</span>
                 </a>
-                <a class="caret-link" href="javascript:;">
+                <a class="caret-link" href="javascript:;" style="width: 20%">
```
NinoRy: Can you increase the link height to 20px (like the link left of it) and keep the caret vertically centered?

modules/Dashboard/Resources/views/timeline/logs.blade.php
```diff
@@ -32,7 +32,7 @@
                 <div class="timeline-content">
                     <h4>
                         <i class="fa {{$bclasses[$log->method]}}"></i>
-                        {{$log->username}} {{$log->method}} <a href="admin/{{$log->model}}/{{$log->model_id}}">{{$log->model}}</a>
+                        {{$log->username}} {{ trans("messages.dashboard_log_$log->method") }} <a href="admin/{{$log->model}}/{{$log->model_id}}">{{ trans("messages.$log->model") }}</a>
```
NinoRy: array syntax would be "messages.dashboard.log.updated"

NinoRy:
Please resolve conflicts with @olebowle

modules/HfcBase/Resources/views/index.blade.php
```diff
@@ -1,7 +1,84 @@
-@extends('HfcBase::layouts.master')
+@extends ('Layout.default')
+
+@section ('income')
+    <h4>{{ \App\Http\Controllers\BaseViewController::translate_view('Net Income', 'Dashboard') }} {{ date('m/Y') }}</h4>
+    <p>
+        @if ($netelements)
+            {{ number_format($netelements['total'], 0, ',', '.') }}
```
NinoRy: there shouldn't be an income section in hfc index.blade

NinoRy:
Pls remove 'Geoposition' of translation "Modem Positionierungsregel Geoposition erstellen" in hfc dashboard quickstart

PR #647, opened 2019-04-09
feature/small-fixes-nino

NinoRy:
#

### DIFF
--- a/modules/BillingBase/Entities/Item.php
+++ b/modules/BillingBase/Entities/Item.php
@@ -91,18 +91,29 @@ class Item extends \BaseModel
         return $ret;
     }
 
+    /**
+     * Get label bootstrap class for colorization
+     *
+     * green: valid
+     * blue:  starts in future
+     * grey:  outdated/expired
+     */
     public function get_bsclass()
     {
-        // Evaluate Colours
-        // green: valid
-        // blue:  starts in future
-        // grey:  outdated/expired
-        // '$this->id' to dont check when index table header is determined!
-        if ($this->id && $this->check_validity('now')) {
+        // Dont check when index table header is determined!
+        if (! $this->id) {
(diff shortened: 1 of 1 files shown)


modules/BillingBase/Entities/Item.php
```diff
@@ -683,10 +681,10 @@ public function creating($item)
 
     public function created($item)
     {
-        // \Log::debug('created item', [$item->id]);
-
         // this is ab(used) here for easily setting the correct values
-        $item->contract->daily_conversion();
+        if (in_array($item->product->type, ['Internet', 'Voip'])) {
```
cschra: Thoughts on this:
why not return inside daily_conversion method, if it is not Internet or voip? -maybe send through a parameter for dirty checking ...
to have only one source of truth and if you want to change product types for execution, you have to change it only in one instead of 3 places

modules/BillingBase/Entities/Item.php
```diff
@@ -91,18 +91,29 @@ public function view_index_label()
...
-        // blue:  starts in future
-        // grey:  outdated/expired
-        // '$this->id' to dont check when index table header is determined!
-        if ($this->id && $this->check_validity('now')) {
+        // Dont check when index table header is determined!
+        if (! $this->id) {
+            return '';
+        }
+
+        if ($this->product->billing_cycle == 'Once' && ! $this->get_end_time() && $this->get_start_time() < strtotime('midnight first day of this month')) {
```
xee8ai: Why check for “&& ! $this->get_end_time()” here?
If cycle is “once” and “start_time” less than this month – shouldn't the billing has been performed regardless of the end date?

xee8ai: Okay, figured it out by myself – that is used for installment…

PR #689, opened 2019-05-24
add new option to open modem edit page in a new tab from the topograp…

aalabed:


schmto:
Thx for this wonderful adaption, i will approve it when
1. styleci issues are fixed:
  https://github.styleci.io/analyses/qJQd9Q?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
You should be able to just push the offered changes towards your tree and PR checks should be green
2. @cschra or @NinoRy  also approve.

PR #691, opened 2019-06-04
Bugfix:  Fix backslash display issue in datatables

cschra:
@Gitrosa found this Bug: when searching for special Characaters a Slash was added, when the search values got pulled from local storage.

This PR, consisting of only one commit, should fix this issue.
It removes backslashes for search data coming from local Storage with a RegEx.


cschra:
@NinoRy what should be the behaviour for #679 ? This PR is fixing datatables stuff and this could is an easy addition.
Currently its 2h -> either update it to be 24 hours or infinite? Maybe discuss with @schmto and @olebowle.

NinoRy:
I would just prefer 24h hours in a first step. Let's see what the feedback is from the customers.

Gitrosa:
@cschra Please take a final look to make sure everything is fine.

PR #698, opened 2019-06-13
Laravel 5.5

cschra:
At first glance, this update will be larger than 5.4, as more of the features we use got updates and changes.

ToDo:
- [x] Bump Composer Dependencies
    - [x] Laravel
    - [x] HTML Package
    - [x] Laravel Modules
- [x] Enable Artisan Command Autoloading (no need to define them in Console Kernel)
- [x] Change Command fire() method to handle(), as fire is no longer supported
- [x] BelongsToMany Relationship changed usage
- [x] Direct array access to models is not supported anymore 
- [x] Look into Cookie Serialization ->Security Bug if App_KEY is leaked (i.e. previous employee)
- [x] Look if we use soft-deleted models and exist
- [x] Look for withCount statements
- [x] Look for usage of Filesystem files method
- [x] Look into request all, has, only and intersect methods and the request() helper itself

``` php
$modem['attributes'];
$modem['original’];
$modem['street'];
$user['name']; 
$modem['attributes']['street'];
// the upper calls are deprecated and don't work anymore
// attributes and original are now private properties
// they can be accessed through following methods
$modem->getAttributes()
$modem->getOriginal()
```


cschra:
Checks for Dunning Module:

- [ ] Change Command fire() method to handle(), as fire is no longer supported
- [ ] BelongsToMany Relationship changed usage
- [ ] Direct array access to models is not supported anymore
- [ ] Look for soft-deleted models and exist
- [ ] Look for withCount statements
- [ ] Look for usage of Filesystem files method
- [ ] Look into request all, has, only and intersect methods and the request() helper itself

cschra:
- [ ] Adjust Codebase for Middleware: ConvertEmptyStringsToNull

database/migrations/2019_07_01_163135_setEmptyStringsToNull.php
```diff
@@ -0,0 +1,117 @@
...
+                    continue;
+                }
+
+                if ($type == 'smallint') {
+                    $type = 'smallInteger';
+                }
+
+                Schema::table($tableName, function (Blueprint $table) use ($column, $nullable, $type) {
+                    if (array_key_exists($column, $nullable)) {
+                        return $table->$type($column)->nullable($nullable[$column])->change();
```
NinoRy: Does it really work with return in this loop?

app/Http/Controllers/BaseController.php
```diff
@@ -232,9 +232,6 @@ protected function prepare_input($data)
                 $field['name'] = str_replace('[]', '', $field['name']);
                 continue; 			// multiselects will have array in data so don't trim
             }
-
-            // trim all inputs as default
-            $data[$field['name']] = trim($data[$field['name']]);
```
NinoRy: This should be kept as users often added whitespaces to e.g. passwords of phonenumbers which led to failed registration.

PR #706, opened 2019-06-21
Update news report with modem count

Gitrosa:


schmto:
even if possibility is <0.5% that this will break anything. Have you test this e.g. demo/production system? in case yes, please feel free to merge this...

PR #708, opened 2019-07-02
Bugfix/update-laravel-5-4

royschn:


royschn:
I've tested it quite a few times of both the job and the command. It is working as far as I can tell.

PR #709, opened 2019-07-02
Feature/small fixes roy

royschn:
#


NinoRy:
Why 2 commits? How does this actually work?

PR #716, opened 2019-07-05
 feature/overdue-debts

NinoRy:
Adaptions to work for Volksbank MT940.sta (Swift) files as well


modules/Dunning/Entities/DefaultTransactionParser.php
```diff
@@ -0,0 +1,504 @@
...
+            $ret = $this->addDebtBySpecialMatch($debt, $sepamandate->contract, $transaction);
+
+            if ($ret) {
+                return true;
+            }
+
+            $hint .= ' '.trans('dunning::messages.transaction.credit.noInvoice.sepa', ['contract' => $sepamandate->contract->number]);
+        }
+
+        if ($hint) {
```
cschra: hint is an empty string anyways... so just add it behind - no need for an "if"

Issue #696, opened 2019-06-07
Modules Dashboards Fixes


NinoRy:
- [x] (1) There are 2 menu entries called 'Voip'. I dont know how the lower one got there, but that entry should be removed. The call data records under this entry dont need to show up in the menu.

- [x] (2a) The timeline shall still be more compact. More like the one in confluence as shown in picture 'timeline.png'. So all actions consecutetively done by one user shall be composited (one "timeline-body") without any redundant information. The user name shall appear on the left side of the timeline and if there are multiple actions the timestamp probably need to appear on the right side together with the action. Just try to implement a layout that is compact and has a good and logical overview/structure!

- [x] (2b) Also the maximum text length of the description containing what changed in one action should be limited - maybe max 2 lines. If it's more it would be nice if it's expandable.

- [x] (3a) The widget with the new tickets shall appear on the global dashboard as well - as it is probably the most important thing to see. I dont know if on top or on the right side (splitted with div 9 to 3 or sth like that). Just implement what you would prefer and we will have a look at it.

- [x] (3b) impaired netelements and impaired services tab shall also be displayed in Global Dashboard

- [x] (4) The user shall be able to select the first shown page after login on it's preferences (where he can set/change his password). It shall be possible to select every module dashboard and also every index page of every model.

- [x] (5) The contract analysis graph needs to be in the billing dashboard

- [x] (6) The widget with the current total contract count is missing. This should be in the provisioning dashboard. Please check the code in the history and just add it again.

olebowle:
(7) Firmware distribution page ("Firmware", admin/modem/firmware) and Impaired Modems (admin/Customer/impaired) are currently gone and should be displayed in Provisioning Dashboard.

cschra:
(8) The News Tab with introduction and current YouTube Video, should be displayed in Base Dashboard
(9) impaired netelements and  impaired services tab is missing should also be displayed in Base Dashboard

xee8ai:
(10)
Most of the models belong to “higher” models: Modem⇒Contract, Item⇒Contract, Mta⇒Modem, Phonenumber⇒Mta, …
So it's impossible, for example, to create a “standalone” phonenumber from dashboard level (this is prevented by the validation rules because of the missing mta_id).

NinoRy:
(10) On top of the general dashboard some kind of internal alert message for the employees shall be configurable - this leads to following sub tasks:
[...]

PR #717, opened 2019-07-10
Feature/modules dashboards 

kokuta:
#696


modules/BillingBase/Resources/views/index.blade.php
```diff
@@ -117,6 +122,74 @@
                     }
                 });
             }
+            // line chart contracts
+            var chart_data_contracts = {!! $contracts_data ? json_encode($contracts_data['chart']) : '{}' !!};
+
+            if (Object.getOwnPropertyNames(chart_data_contracts).length != 0) {
+
```
NinoRy: You just added these lines. Shouldn't they be removed in modules/ProvBase/Resources/views/index.blade.php or are they still needed there?

kokuta: Yes you are right, we no longer have contracts data chart on Provbase dashboard.

NinoRy:
- Impaired modems and Firmware distribution is in global Dashboard, but must be in Provisioning Dashboard.
- @section ('weekly_contracts') in Provisioning Dashboard shall be in Billing Dashboard

PR #720, opened 2019-07-16
feature/small-fixes-nino

NinoRy:


modules/BillingBase/Entities/SettlementRun.php
```diff
@@ -92,9 +92,9 @@ public function set_index_delete()
         return $this->index_delete_disabled;
     }
 
-    public function created_at_toDateString()
+    public function executed_at_toDateString()
     {
-        return $this->created_at->toDateString();
+        return $this->executed_at->toDateString();
```
cschra: this can be done natively by laravel with property $casts -> see [here](https://laravel.com/docs/5.8/eloquent-mutators#attribute-casting)

PR #733, opened 2019-07-25
feature/small-fixes-nino

NinoRy:


modules/BillingBase/Entities/SettlementRun.php
```diff
@@ -567,36 +567,52 @@ private function getSepaAccounts($account = null)
...
+     *  Used to clean up directory and DB on repeated run of SettlementRun::execute()
+     *
+     * @param obj   Delete only invoices related to specific SepaAccount, 0 - delete all invoices of this SR
+     * @param bool  Delete only temporary LaTeX files created during SettlementRun and check if Invoices where created successfully
      */
-    public function delete_current_invoices($sepaaccount)
+    public function delete_current_invoices($sepaaccount, $tempFiles = false)
     {
-        $query = Invoice::whereBetween('created_at', [date('Y-m-01 00:00:00'), date('Y-m-01 00:00:00', strtotime('next month'))]);
+        $query = \DB::table('invoice')->whereBetween('created_at', [date('Y-m-01'), date('Y-m-01', strtotime('next month'))]);
```
cschra: Is "where deleted_at is null" necessary? Because the table uses soft deletes, but does this even occur?

NinoRy: must not be used

PR #741, opened 2019-07-31
Bugfix/small fixes nino

NinoRy:


app/Http/Controllers/BaseViewController.php
```diff
@@ -531,6 +531,7 @@ public static function view_main_menus() : array
         $globalPages = Config::get('base.'.$configMenuItemKey);
 
         $menu['Global']['link'] = Config::get('base.link');
+        $menu['Global']['translated_name'] = 'Global';
```
cschra: should be translated!? As in Spanish it might not be "Global"

PR #756, opened 2019-08-06
Bugfix/small fixes nino

NinoRy:

### DIFF
--- a/app/BaseModel.php
+++ b/app/BaseModel.php
@@ -395,7 +395,7 @@ class BaseModel extends Eloquent
     }
 
     /**
-     * Get all models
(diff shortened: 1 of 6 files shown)


modules/ProvBase/Entities/Contract.php
```diff
@@ -1389,6 +1389,13 @@ public function updated($contract)
             }
         }
     }
+
+    public function deleting($contract)
+    {
+        if ($contract->cccUser) {
```
cschra: I think it would be better to use a SQL foreign constraint for this -> this way, the database handles this task automatically and we keep the PHP simple

PR #759, opened 2019-08-12
Feature/small fixes nino

NinoRy:


modules/HfcCustomer/Http/Controllers/CustomerTopoController.php
```diff
@@ -241,41 +240,64 @@ public function show_topo($modems, $row = null, $allModels = null)
...
-        return \View::make('HfcBase::Tree.topo', $this->compact_prep_view(compact('file', 'target', 'route_name', 'view_header', 'body_onload', 'modems', 'tabs', 'kmls', 'models')));
+        return \View::make('HfcBase::Tree.topo', $this->compact_prep_view(compact('file', 'target', 'route_name', 'view_header', 'body_onload', 'tabs', 'kmls', 'models')));
+    }
+
+    /**
+     * Get Query for all Modems belonging to a valid contract or being online (us_pwr > 0)
+     *
+     * @return obj Illuminate\Database\Query\Bilder
+     */
+    private function getModemBaseQuery()
```
cschra: this could be a scope :wink:  -> rename function to scope[name]($query) and utilize it with Modem::[name]

resources/views/Generic/index.blade.php
```diff
@@ -68,6 +68,12 @@
     {{-- database entries inside a form with checkboxes to be able to delete one or more entries --}}
     {{ Form::open(array('route' => array($route_name.'.destroy', 0), 'method' => 'delete', 'id' => 'IndexForm')) }}
     {{-- INIT DT --}}
+
+    @php
+        $methodExists = method_exists(BaseController::get_model_obj(), 'view_index_label');
```
cschra: why not:
```suggestion
        $methodExists = method_exists($model, 'view_index_label');
```
?

cschra: The rest of the refactor of the index blade looks very good - very good simplification and reduction of function calls!

modules/HfcCustomer/Http/Controllers/CustomerTopoController.php
cschra: [Laravel Doku](https://laravel.com/docs/5.6/eloquent#local-scopes)

NinoRy: I think this won't work without Eloquent, does it?

cschra: This is the nomal Querybuilder... so I would assume it is also working without hydrating eloquent modes and your use case seems to be perfect for queryscoping... basically you just put a nametag on queries you often use...

cschra: you are right DB table does not support scopes. valid point.
So your implementation seems to be the desired implementation.
will resolve this and approve the PR

cschra: You are right... this is only beneficial when using eloquent

modules/BillingBase/Database/Migrations/2019_09_24_001000_changeFloatToDecimal.php
```diff
@@ -0,0 +1,47 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('item', function (Blueprint $table) {
+            $table->decimal('credit_amount', 10, 4)->nullable()->change();
+        });
+
+        // Change enums to string because of Laravel DBAL bug: https://stackoverflow.com/questions/29165259/laravel-db-migration-renamecolumn-error-unknown-database-type-enum-requested/32860409#32860409
+        DB::statement('ALTER TABLE product MODIFY COLUMN type varchar(50) NOT NULL');
```
cschra: How does this affect the getPossibleEnumValues() Funktion? Does this still work?

cschra: to work with enums the following workaround is also possible:
```php
    public function __construct()
    {
        DB::getDoctrineSchemaManager()->getDatabasePlatform()->registerDoctrineTypeMapping('enum', 'string');
        Schema::defaultStringLength(191);
    }
```
maybe we should considder adding this to the BaseMigration class

NinoRy: Ah yes. Haven't thought about this function.
Yes. I think adding the register function is the best way. Thanks!

PR #760, opened 2019-08-14
Bugfix/make envia related phonenumbers deletable

xee8ai:
#


xee8ai:
Phonenumbers with canceled EnviaContract are now deletable.

Also refactored some things I noticed on the way.

modules/ProvVoip/Entities/PhonenumberManagement.php
```diff
@@ -221,7 +221,7 @@ public function delete()
                     return false;
                 }
                 if (in_array($this->envia_contract->state, ['Gekündigt'])) {
-                    if ($this->envia_contract->end_date <= \Carbon\Carbon::now()->toDateTimeString()) {
+                    if ($this->envia_contract->end_date > \Carbon\Carbon::now()->toDateTimeString()) {
```
cschra: naravel has the helper now() built in, so no need to call the Carbon class.

modules/ProvVoip/Entities/PhonenumberManagement.php
```diff
@@ -207,38 +207,35 @@ public function delete()
...
             // check if there is a not completely terminated envia TEL contract related to this management
             if ($this->envia_contract) {
                 if (in_array($this->envia_contract->state, ['Aktiv', 'In Realisierung'])) {
                     $msg = "Cannot delete PhonenumberManagement $this->id: There is an active or pending envia TEL contract";
-                    if (\Str::endsWith($prev, 'edit')) {
-                        \Session::push('tmp_error_above_relations', $msg);
-                    } else {
-                        \Session::push('tmp_error_above_index_list', $msg);
-                    }
+                    \Session::push($msg_target, $msg);
```
cschra: "use Session" and maybe also "Illuminate/Support/Str" as its used more than once...

app/BaseModel.php
```diff
@@ -988,6 +988,74 @@ public function check_validity($timespan = 'monthly', $time = null, $start_end_t
...
+            if (!in_array($place, $allowed_places)) {
+                throw new \UnexpectedValueException('$place has to be in ['.implode('|', $allowed_places).'], “'.$place.'” given.');
+            };
+        };
+
+        // build the message target
+        $target = 'tmp_'.$type.'_above_'.$place;
+
+        // push to session ⇒ will be shown once via resources/views/Generic/above_infos.blade.php
+        try {
```
cschra: no try catch needed - just do 
```php
if (! Session::has($target)) { //this is not tested, just my assumption how it could work
     \Log::error()
    return false
}

\Session::push($target, $msg);
return true;
```
And then again: when does the session push fail? This would be Laravels responsibility to report that error, so a simple 
```php
\Session::push($target, $msg);
return true;
```
would do the rrick in my Opinion

app/BaseModel.php
```diff
@@ -988,6 +988,74 @@ public function check_validity($timespan = 'monthly', $time = null, $start_end_t
...
+     * @param string    $place  Where to show the message above [index_list, form, relations];
+     *                              if not given try to determine from previous URL
+     *
+     * @return boolean  true if message could be generated, false else
+     *
+     * @author Patrick Reichel
+     */
+    public function addAboveMessage($msg, $type='info', $place=null) {
+
+        $allowed_types = [
```
cschra: method name camel case and variable names are snake case?
TODO: which style do we want to use? This comment is here to avoid even more fragmentation
I would go with all camel or all snake, but not mixing them

app/Http/Controllers/BaseController.php
```diff
@@ -638,8 +638,8 @@ public function store($redirect = true)
         if ($validator->fails()) {
             Log::info('Validation Rule Error: '.$validator->errors());
 
-            $msg = 'Input invalid – please correct the following errors';
-            Session::push('tmp_error_above_form', $msg);
+            $msg = trans('validation.invalid_input');
```
cschra: on simple messages like here I would inline the msg variable
-> as its just one expression without parameters (there are a few more places where this applies)

modules/BillingBase/Entities/Product.php
```diff
@@ -113,44 +113,21 @@ public function costcenter()
...
      *
      * @return bool
      */
     public function delete()
     {
         // only delete if there are no items assigned for this product
         if ($this->item()->count() > 0) {
-            Session::push('tmp_error_above_index_list', trans('messages.assigned_items'));
+            $msg = trans('messages.assigned_items');
+            $this->addAboveMessage($msg, 'error');
```
cschra: i.e. here ... (inline)

modules/ProvVoipEnvia/Resources/lang/de/messages.php
```diff
@@ -50,7 +50,10 @@
     'description' => 'Beschreibung',
     'differentActivationDates' => 'Unterschiedliche Startdaten für die übergebenen Rufnummern (:0, :1)',
     'differentSubscriberData' => 'Differences in subscriber data (:0 != :1)',
+    'doManuallyNow' => 'DAS JETZT MANUELL ERLEDIGEN!',
```
cschra: sounds weird ... in my opinion "JETZT MANUELL ERLEDIGEN!" sounds more natural  (without context... might fit into context)
and why is all in Capital letters? Do we need to shout at the user?

modules/ProvVoipEnvia/Resources/lang/de/messages.php
```diff
@@ -70,6 +73,7 @@
     'mgcpNotImplemented' => 'MGCP wird nicht unterstützt.',
     'miscGetKeysUnused' => 'Achtung: Diese Daten werden derzeit nicht zur Aktualisierung unserer Datenbank genutzt.',
     'miscGetKeysWarning' => 'Achtung: Die Daten für diese Schlüsselwerte sollen nicht öfter als einmal täglich heruntergeladen werden.',
+    'modemStillNumbersAttached' => 'Dem Modem :href sind noch weitere Rufnummern zugeordnet! Diese sollten ebenfalls umgehängt werden: :numbers',
```
cschra: umgehängt? -> portiert?

modules/ProvVoipEnvia/Resources/lang/de/messages.php
```diff
@@ -84,7 +88,10 @@
     'no_management' => 'Kein PhonenumberManagement',
     'orderCancelFailed' => 'Stornierung fehlgeschlagen. envia-TEL-Auftrag :0 (envia-TEL-Auftrags-ID :1) wieder hergestellt.',
     'orderCancelSuccess' => 'envia-TEL-Auftrag :0 (envia-TEL-Auftrags-ID :1) wurde storniert.',
+    'orderCannotCreateIdMissing' => 'envia-TEL-Auftrag kann nicht angelegt werden – weder contract_id noch modem_id noch phonenumbermanagement_id wurde übergeben.',
```
cschra: *wurden

modules/ProvVoipEnvia/Resources/lang/de/messages.php
```diff
@@ -84,7 +88,10 @@
...
+    'orderCannotCreateModelMissing' => 'envia-TEL-Auftrag kann nicht angelegt werden – :model :id existiert nicht',
```
cschra: :model mit der ID :id

modules/ProvVoipEnvia/Resources/lang/de/messages.php
```diff
@@ -123,6 +137,7 @@
     'sendToEnviaNow' => 'Ja, Daten jetzt schicken!',
     'settingExternalCustId' => 'Setze envia-TEL-Kundenreferenz für Vertrag :0 auf :1.',
     'sipChanged' => 'SIP-Daten für Rufnummer :0 aktualisiert.',
+    'sipDateNotChangedAutomaticallyAtEnvia' => 'Die Änderung der SIP-Daten wird nicht automatisch an envia TEL übermittelt.<br>Sie müssen :href',
```
cschra: is there something missing?

modules/ProvVoipEnvia/Resources/views/EnviaContract/edit.blade.php
```diff
@@ -2,6 +2,12 @@
 @extends('Generic.edit')
 
 @section('content_left')
+    @include ('Generic.logging')
+
+    <?php
```
cschra: I would recommend to use the @php Blade drective
@php
    $blade_type = 'form';
@endphp

app/BaseModel.php
NinoRy: Yes, all camel case is the new convention

modules/ProvVoip/Entities/PhonenumberManagement.php
xee8ai: Accepted. Now using now() :-)

app/BaseModel.php
xee8ai: Removed.

app/Http/Controllers/BaseController.php
xee8ai: IMO my version is better readable compared to nested function calls.

modules/ProvVoipEnvia/Resources/lang/de/messages.php
xee8ai: No :-)
:href contains a complete sentence.

PR #763, opened 2019-08-18
Feature/modems cmts support checks

kokuta:
#


kokuta:
Tests need on live running system with connected modems and cmts. 
As I made some isolated tests, maybe would be good if we test on a live integrated system, and see if these OID return the desired value/s.
Run `php artisan nms:hardware-support` to run the command

modules/ProvBase/Database/Migrations/2019_05_29_014631_add_field_support_state_to_modem_table.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('modem', function (Blueprint $table) {
+            $table->enum('support_state', ['verifying', 'full-support', 'not-supported', 'restricted'])->default('verifying');
```
cschra: avoid enum, as doctrine has issues processing those. We recommend to create text columns and restrict the content via validation rules.

modules/ProvBase/Console/HardwareSupportCommand.php
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Execute the console command - Check for hardware support for each modem and each CMTS modules
+     *
+     * @return mixed
+     */
+    public function fire()
+    {
+
+        $this->snmp_def_mode();
+        $modems = Modem::all();
```
cschra: this is very performance intensive as eloquent hydrates all the models. Consider using DB::table or Scope your query.

modules/ProvBase/Console/HardwareSupportCommand.php
```diff
@@ -0,0 +1,131 @@
...
+                }
+            }
+            $modem_serial_no_md5 = md5($modem->serial_num);
+            $contents = file_get_contents('https://support.nmsprime.com/hwsn/api.php?q=' . $modem_serial_no_md5);
+
+            if ($contents !== '') {
+                $result = json_decode($contents, true);
+                if (isset($result[$modem_serial_no_md5]) && $result[$modem_serial_no_md5] === 'valid') {
+                    $support_state = 'full-supported';
+                } elseif ((Carbon::parse($modem->created_at))->diffInWeeks(Carbon::now()) < 6) {
```
cschra: as far as I know "created_at" is always cast to a Carbon Date by Laravel and there is also a shorthand helper now()
so this line could be simplified to
```suggestion
                } elseif (($modem->created_at->diffInWeeks(now())) < 6) {
```

modules/ProvBase/Console/HardwareSupportCommand.php
```diff
@@ -0,0 +1,131 @@
...
+                }
+
+                if ($count_found) {
+                    $percentage = $count_found / count($cmts_serials) * 100;
+
+                    if ($percentage > 95) {
+                        $support_state = 'full-supported';
+                    } elseif ($percentage > 80 && $percentage <= 95) {
+                        $support_state = 'restricted';
+                    } elseif ((Carbon::parse($cmts->created_at))->diffInWeeks(Carbon::now()) < 6) {
```
cschra: same as above...

modules/ProvBase/Console/HardwareSupportCommand.php
```diff
@@ -0,0 +1,131 @@
...
+     */
+    public function handle()
+    {
+        $this->snmp_def_mode();
+        $modems = DB::table('modem')->get();
+        $cmtses = DB::table('cmts')->get();
+        $ro_community = ProvBase::first()->ro_community;
+
+        foreach ($modems as $modem) {
+            $hostname = $modem->hostname . '.' . $this->domain_name;
```
cschra:
```suggestion
            $hostname = "{$modem->hostname}.{$domain_name}";
```

modules/ProvBase/Console/HardwareSupportCommand.php
```diff
@@ -82,7 +82,7 @@ public function handle()
             //TODO: Test snmpwalk response a live system, not tested on a fully integrated system
             try {
                 $cmts_serials = snmpwalk($hostname, $ro_community, '1.3.6.1.2.1.47.1.1.1.1.11'); //TODO: Handle Exception
-
+                var_dump($cmts_serials);
```
cschra: hint FYI: you can use laravels "dump()" or "dd()" helper OR our own "d()" function ;-)

modules/ProvBase/Console/HardwareSupportCommand.php
```diff
@@ -0,0 +1,131 @@
...
+    protected $domain_name = '';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        $this->domain_name = ProvBase::first()->domain_name;
```
cschra:
```suggestion
        $this->provBaseSettings = ProvBase::first(); //saves DB queries
```

modules/ProvBase/Console/HardwareSupportCommand.php
```diff
@@ -0,0 +1,131 @@
...
+     * Execute the console command - Check for hardware support for each modem and each CMTS modules
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $this->snmp_def_mode();
+        $modems = DB::table('modem')->get();
+        $cmtses = DB::table('cmts')->get();
+        $ro_community = ProvBase::first()->ro_community;
```
cschra:
```suggestion
        $ro_community = $this->provBaseSettings->ro_community;
        $domain_name = $this->provBaseSettings->domain_name;
```

modules/ProvBase/Console/HardwareSupportCommand.php
kokuta: Yes you are right if we would get the records as eloquent model instances eg: Modem::all(().
In case of retrieving them as collections, the date fields are not wrapped as Carbon instance hence the "Call to a member function diffInWeeks() on string" error would occur.

cschra: sure :man_facepalming: I forgot about that.

then no objections :grinning:

cschra:
Code is okay from my point of view, but let @NinoRy have a glance at it.
You have some duplicate commits. It would be great if you could rebase (interactively) to cleanup the history to make it more comprehensible

resources/lang/en/dt_header.php
```diff
@@ -178,6 +179,7 @@
     'modem.street' => 'Street',
     'modem.sw_rev' => 'Firmware Version',
     'modem.us_pwr' => 'US level',
+    'modem.formatted_support_state' => 'Suport State',
```
NinoRy: Missspelled

app/Console/Kernel.php
```diff
@@ -146,6 +146,9 @@ protected function schedule(Schedule $schedule)
 
             // update firmware version + model strings of all modems once a day
             $schedule->call('\Modules\ProvBase\Entities\Modem@update_model_firmware')->daily();
+
+            // Hardware support check for modems and CMTS
+            $schedule->command('nms:hardware-support daily')->twiceDaily(10, 14);
```
NinoRy: Only execute on valid SLA

PR #772, opened 2019-09-12
Update request of news.json

Gitrosa:


modules/Dashboard/Http/Controllers/DashboardController.php
```diff
@@ -175,17 +175,33 @@ public function newsLoadToFile()
...
+            ->where(function ($query) {
+                $query->where('valid_to', null)
+                    ->orWhere('valid_to', '>=', date('Y-m-d'));
+            })
+            ->count();
         }
 
         $files = [
-            'news.json' => "$support/news.php?ns=".urlencode($sla->get_sla_size()).'&sla='.urlencode($sla->name).'&mc='.$modemcount,
+            'news.json' => "$support/news.php?ns=&sla=".urlencode(\App\SLA::first()->name).'&mc='.$numModems.'&nm='.$numCmts.'&nn='.$numNetelements.'&nt='.$numTvbillings,
```
cschra: did you test this? 
There might be a Typo as there is no "App\SLA" class -> "App\Sla"

cschra: Laravel seems to be creating something like a realtime alias. If you provide the full namespace the classnames are case insensitive. However I would recommend sticking to the original classname.
Laravel magic never ceases to amaze me :open_mouth:

schmto: @cschra : since @Gitrosa is on holiday, all that i noticed is, that this code is tested. So i will approve this without merging. I would recommend @Gitrosa to answer @cschra issues and if we have 100% clarification: feel free to merge..

schmto: maybe one question: will there always be a valid \App\SLA::first() entry for all NMS Prime instances? otherwise this should be tested.

cschra: > maybe one question: will there always be a valid \App\SLA::first() entry for all NMS Prime instances? otherwise this should be tested.

I am not 100% sure what you mean:
1. The first "Sla" model is created with the migration - so `Sla::first()` will never return `null`
2. Due to the Laravel "Magic" described above it `SLA::first()`, `sla::first()`, `sLa::first()` or any other combination will work.

Gitrosa: The Sla class isn't needed for this, but is as far as I know used somewhere else for the supportticket-form.
PHP is case-insensitive for classnames, so that works anyways but as @cschra said, we should be sticking to the original class name.

PR #775, opened 2019-09-13
 Introducing Speed test feature to Customers Center

aalabed:
#


modules/Ccc/Resources/views/index.blade.php
```diff
@@ -61,6 +59,304 @@ class="btn btn-xs btn-icon btn-circle btn-warning d-flex"
       </div>
     @endforeach
   </div>
+
+ <div class="row">
+      <div class="col-md-6 ui-sortable">
+        <div class="panel panel-inverse card-2 d-flex flex-column">
+          <div class="panel-heading d-flex flex-row justify-content-between">
+            <h4 class="panel-title d-flex">
```
NinoRy: Please include a separate blade 'speedtest.blade.php' here and outsource code

NinoRy:
Hey @kokuta,
can you have a look at the conflicts here? Does it make sense to keep ".tab-overflow .nav.nav-pills > li" twice (line 645 and 86,87) and is it necessary to use the !important rules?

public/components/nmsprime.css
```diff
@@ -530,3 +530,98 @@ table .index_check {
 #page-container.page-sidebar-minified .nav ul.sub-menu li i.fa+span{
   display: inline;
 }
+
+
+body{
+    text-align:center;
+    font-family:"Roboto",sans-serif;
+  }
```
kokuta: Did you intentionally made this alignment? This will affect all the pages. I already noticed some center alignment where it shouldn't. I would suggest to use bootstraps classes  (eg: text-center), or create your own if want specific styling.

kokuta:
> Hey @kokuta,
> can you have a look at the conflicts here? Does it make sense to keep ".tab-overflow .nav.nav-pills > li" twice (line 645 and 86,87) and is it necessary to use the !important rules?

Yes you are right, it does not make sense to keep ".tab-overflow .nav.nav-pills  li" twice. 
As I checked, it is not necessary also to use the !important rules.

public/components/nmsprime.css
NinoRy: @aalabed Do you know why you added this?

cschra:
@aalabed please rebase your branch to fix style issues and ut your Javascript inside the created Javascript section

modules/Ccc/Resources/views/index.blade.php
```diff
@@ -87,3 +87,129 @@ class="btn btn-xs btn-icon btn-circle btn-warning d-flex"
 
 @stop
 
+@section('java script')
+<script type="text/javascript" src="speedtest.js"></script>
+<script type="text/javascript">
```
NinoRy: When you move it it should be deleted somewhere!?

modules/Ccc/Resources/views/speed_test.blade.php
```diff
@@ -0,0 +1,187 @@
...
+  drawMeter(I("ulMeter"),0,meterBk,ulColor,0);
+  drawMeter(I("pingMeter"),0,meterBk,pingColor,0);
+  drawMeter(I("jitMeter"),0,meterBk,jitColor,0);
+  I("dlText").textContent="";
+  I("ulText").textContent="";
+  I("pingText").textContent="";
+  I("jitText").textContent="";
+  I("ip").textContent="";
+}
+</script>
```
cschra: this should be deleted...

PR #779, opened 2019-09-23
Bugfix L5.5 Tree View

cschra:
set parent_id to NULL if there is no parent


cschra:
@olebowle should merge as icinga/cacti adjustments are necessary after applying this

PR #792, opened 2019-10-23
bugfix/small-fixes-nino

NinoRy:


modules/OverdueDebts/Resources/lang/de/messages.php
```diff
@@ -7,6 +7,7 @@
     'csvImportActive' => 'Der OP-Import läuft gerade. Bitte warten Sie bis der Prozess beendet wurde.',
     'import' => [
         'block' => 'Blockiere Internetzugriff und füge Mahngebühren hinzu...',
+        'columnCountError' => 'Die Anzahl der Spalten in der CSV ist fehlerhaft. Die Anzahl muss 16 Spalten betragen.',
```
cschra: number is static and should be changed to :number as well

app/Console/Commands/SyncLanguagesCommand.php
```diff
@@ -47,7 +47,7 @@ public function handle()
 
         foreach ($languagefiles as $languagefile) {
             $originalPath = base_path('resources/lang/en/').$languagefile;
-            $originalArray = collect(require($originalPath));
+            $originalArray = collect(require $originalPath);
```
cschra: These StyleCI issues are also fixed by @olebowle and already in dev I think - could you rebase to dev and check if they still appear

NinoRy: done

modules/OverdueDebts/Resources/lang/de/messages.php
NinoRy: right ... done

PR #802, opened 2019-11-25
Bugfix: Fix wrong request calls.

cschra:
#Fix some leftover bugs of Laravel Upgrade and moving to the Request-Facade

### DIFF
--- a/modules/ProvBase/Http/Controllers/ConfigfileController.php
+++ b/modules/ProvBase/Http/Controllers/ConfigfileController.php
@@ -157,6 +157,18 @@ class ConfigfileController extends \BaseController
         return str_replace(array_shift($parentIds[0]), 'parent_id":'.Request::get('parent_id').',', $content);
     }
 
+    /**
+     * If the Imported Configfiles have IDs that are in the range of the new
+     * created Configfiles the IDs are overwritten and cause DB errors. The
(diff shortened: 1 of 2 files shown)


modules/ProvBase/Http/Controllers/ConfigfileController.php
```diff
@@ -194,22 +232,23 @@ public function recreateTree($content, $hasName, $originalConfigfiles)
      *
      * @author Roy Schneider
      * @param array $content
```
NinoRy: the param for this comment has changed!

cschra: fixed in ddea3c4

NinoRy: still not fixed!

modules/ProvBase/Http/Controllers/ConfigfileController.php
```diff
@@ -194,22 +235,23 @@ public function recreateTree($content, $hasName, $originalConfigfiles)
...
```
NinoRy:
```suggestion
     * @param array $configfile
```

cschra: fixed in ecb76a0

PR #805, opened 2019-12-04
Bugfix/small fixes nino

NinoRy:
#

### DIFF
--- a/modules/ProvBase/Http/Controllers/ContractController.php
+++ b/modules/ProvBase/Http/Controllers/ContractController.php
@@ -208,7 +208,7 @@ class ContractController extends \BaseController
     {
         if (\Module::collections()->has('PropertyManagement')) {
             // Only group contracts without modems can belong to a Realty directly - with CMs realty_id must be null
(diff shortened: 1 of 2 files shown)


modules/ProvBase/Http/Controllers/ContractController.php
```diff
@@ -208,7 +208,7 @@ public function prepare_rules($rules, $data)
     {
         if (\Module::collections()->has('PropertyManagement')) {
             // Only group contracts without modems can belong to a Realty directly - with CMs realty_id must be null
-            if ($data['realty_id'] && isset($data['id'])) {
+            if (isset($data['realty_id']) && $data['realty_id'] && isset($data['id'])) {
```
cschra: Why is this needed anyways? $data['realty_id'] is not used here

NinoRy: There is a sense behind, but it's not worth to explain it, as it will be removed by the propertymanagement branch anyway

PR #806, opened 2019-12-05
Feature/property management

NinoRy:
#

### DIFF
--- a/modules/PropertyManagement/Entities/Contact.php
+++ b/modules/PropertyManagement/Entities/Contact.php
@@ -66,6 +66,9 @@ class Contact extends \BaseModel
         $ret['Edit']['Realty']['class'] = 'Realty';
         $ret['Edit']['Realty']['relation'] = $this->realties;
 
+        $ret['Edit']['GroupContracts']['class'] = 'Contract';
+        $ret['Edit']['GroupContracts']['relation'] = $this->contracts;
+
         return $ret;
     }
 
@@ -75,6 +78,11 @@ class Contact extends \BaseModel
     public function realties()
     {
         return $this->hasMany(Realty::class);
+    }
+
+    public function contracts()
+    {
+        return $this->HasMany(\Modules\ProvBase\Entities\Contract::class);
     }
 
     public static function labelFromData($contact = null)
--- a/modules/PropertyManagement/Entities/Contact.php
+++ b/modules/PropertyManagement/Entities/Contact.php
@@ -82,7 +82,7 @@ class Contact extends \BaseModel
 
     public function contracts()
     {
-        return $this->HasMany(\Modules\ProvBase\Entities\Contract::class);
+        return $this->hasMany(\Modules\ProvBase\Entities\Contract::class);
     }
 
     public static function labelFromData($contact = null)
--- a/modules/PropertyManagement/Entities/Realty.php
+++ b/modules/PropertyManagement/Entities/Realty.php
@@ -54,7 +54,7 @@ class Realty extends \BaseModel
     {
         $bsclass = 'success';
 
-        if ($this->group_contract) {
+        if ($this->contract_id) {
             $bsclass = 'warning';
         } elseif ($this->concession_agreement) {
             $bsclass = 'info';
@@ -68,7 +68,7 @@ class Realty extends \BaseModel
                 'index_header' => ['id', "$this->table.name", 'number', 'street', 'house_nr', 'zip', 'city',
                     "$this->table.contact_id", "$this->table.contact_local_id",
                     'expansion_degree', "$this->table.concession_agreement",
-                    "$this->table.agreement_from", "$this->table.agreement_to", "$this->table.last_restoration_on", 'group_contract',
+                    "$this->table.agreement_from", "$this->table.agreement_to", "$this->table.last_restoration_on",
                     "$this->table.apartmentCountConnected", "$this->table.apartmentCount",
                     ],
                 'header' => $label,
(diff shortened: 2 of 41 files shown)


modules/ProvBase/Entities/Contract.php
```diff
@@ -400,12 +395,34 @@ public function apartment()
...
+
+        $realty = $this->realty();
+        $this->setRelation('realty', $realty);
+
+        return $realty;
+    }
+
+    public function realty()
+    {
+        return self::join('modem', 'modem.contract_id', 'contract.id')
```
cschra: this should be reviseted when we surpass Laravel 5.8 as there is support for [Has many through](https://laravel.com/docs/5.8/eloquent-relationships#has-one-through) and [intermediate tables](https://laravel.com/docs/5.8/eloquent-relationships#defining-custom-intermediate-table-models)

modules/PropertyManagement/Entities/Contact.php
```diff
@@ -77,6 +80,11 @@ public function realties()
         return $this->HasMany(Realty::class);
     }
 
+    public function contracts()
+    {
+        return $this->HasMany(\Modules\ProvBase\Entities\Contract::class);
```
cschra: as already discussed we should use the came case version of the relationship, see [Laravel Docs - hasMany](https://laravel.com/docs/5.8/eloquent-relationships#one-to-many) and be consistent throughout the project...

modules/ProvBase/Entities/Contract.php
```diff
@@ -384,9 +391,9 @@ public function tickets()
         return $this->hasMany('Modules\Ticketsystem\Entities\Ticket');
     }
 
-    public function realty()
+    public function realties()
     {
-        return $this->belongsTo(\Modules\PropertyManagement\Entities\Realty::class);
+        return $this->HasMany(\Modules\PropertyManagement\Entities\Realty::class);
```
cschra: here as well

modules/PropertyManagement/Http/Controllers/CutoffListController.php
```diff
@@ -35,16 +33,22 @@ public function index_datatables_ajax()
...
+
+            if (strtotime($object->contract_end) < strtotime('-8 week')) {
+                $bsclass = 'warning';
+            }
+
+            if (strtotime($object->contract_end) < strtotime('-12 week')) {
+                $bsclass = 'danger';
+            }
+
+            return $bsclass;
```
cschra:
```php
            if (strtotime($object->contract_end) < strtotime('-8 week')) {
                return 'warning';
            }

            if (strtotime($object->contract_end) < strtotime('-12 week')) {
                return 'danger';
            }

           return 'info';
```
or
```php
            if ($object->contract_end->subWeeks(8)->isBefore(now())) {
                return 'warning';
            }

            if ($object->contract_end->subWeeks(12)->isBefore(now())) {
                return 'danger';
            }

           return 'info';
```
?

modules/PropertyManagement/Entities/Contact.php
NinoRy: fixed in 1e6b80a31

modules/ProvBase/Entities/Contract.php
NinoRy: fixed in 1e6b80a31

modules/PropertyManagement/Entities/Realty.php
```diff
@@ -211,20 +178,22 @@ public function localContact()
      * @param bool $withGroupContract, $withModems
      * @return Illuminate\Database\Eloquent\Collection of \Modules\ProvBase\Entities\Contract
      */
-    public function getRelatedContracts($withGroupContract, $withModems = false)
+    public function getApartmentContracts($withGroupContract, $withModems = false)
     {
-        $contracts1 = \Modules\ProvBase\Entities\Contract::join('modem as am', 'am.contract_id', 'contract.id')
-            ->join('apartment', 'am.apartment_id', 'apartment.id')
+        if ($this->apartments->isEmpty()) {
+            return collect();
```
cschra: I noticed this -> this returns a illuminate\support\Collection and NOT an Eloquent\Collection as denoted in the DocBlock. 
to get an Eloquent Collection the code has to be
```php
use \Modules\ProvBase\Entities\Contract;
```
``` php
if ($this->apartments->isEmpty()) {
    return (new Contract)->newCollection();
}
```

resources/lang/de/view.php
```diff
@@ -1,6 +1,8 @@
 <?php
 
 return [
+        'false' => 'ja',
+        'true' => 'nein',
```
cschra: this is a bit counter intuitive? is this correct?

modules/ProvBase/Entities/Contract.php
cschra: Could you please leave a TODO Laravel 5.8 comment or something like this, so that I can grep these when updating. Thx

modules/PropertyManagement/Http/Controllers/CutoffListController.php
cschra: We can leave it as is, but just as a side note: Laravel creates a new independent Carbon instance on cast -> so subWeeks would work as expected...

resources/lang/de/view.php
NinoRy: lol. thx. This is of course wrong

modules/PropertyManagement/Entities/Realty.php
NinoRy: changed in af7d5acdf

modules/ProvBase/Entities/Contract.php
NinoRy: added in af7d5acdf

PR #823, opened 2020-01-07
Make salutation and academic degree options configurable.

xee8ai:
#


modules/ProvBase/Database/Migrations/2020_01_07_000100_create_salutation_and_academic_degree_files.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+
+class CreateSalutationAndAcademicDegreeFiles extends Migration
```
olebowle: - Our migrations should always extend BaseMigration
- I would consider this migration an install migration, thus it should start with install in both filename and class name

PR #837, opened 2020-01-16
Small fixes and adaptions

royschn:
#


royschn:
#807

modules/BillingBase/Database/Migrations/2020_01_16_103046_addDeprecatedProductColumn.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table($this->tablename, function (Blueprint $table) {
+            $table->boolean('deprecated')->default(NULL);
```
NinoRy: default(NULL) should make no sense when column is not nullable. And i would keep it as not nullable for boolean values. Then it's assured that only 0 and 1 values are possible and not a third value (NULL).

resources/lang/de/helper.php
```diff
@@ -54,6 +54,7 @@
         'pod'                   => 'Beispiele: 14D (14 Tage), 3M (Drei Monate), 1Y (Ein Jahr)',
         'proportional'          => 'Setzen Sie diesen Haken, wenn Posten, die innerhalb des aktuellen Abrechnungszyklus beginnen, anteilig berechnet werden sollen. Somit würde bei einem monatlich abzurechnenden Produkt mit Beginn in der Mitte des Monats im aktuellen Abrechnungszyklus nur die Hälfte des vollen Preises abgerechnet werden.',
         'Type'                  => 'Alle Felder außer dem Abrechnungszyklus müssen vor einer Änderung des Produkts gelöscht werden! Andernfalls können die Produkte in den meisten Fällen nicht gespeichert werden.',
+        'deprecated'            => 'Setzten Sie diesen Haken, wenn das Produkt veraltet ist. Dadurch kann es nicht mehr beim Erstellen/Ändern von Posten ausgewählt werden.',
```
NinoRy:
```suggestion
        'deprecated'            => 'Setzen Sie diesen Haken, wenn das Produkt veraltet ist. Dadurch kann es nicht mehr beim Erstellen/Ändern von Posten ausgewählt werden.',
```

resources/lang/en/helper.php
```diff
@@ -54,6 +54,7 @@
         'pod'                   => 'E.g. 14D (14 days), 3M (three months), 1Y (one year)',
         'proportional'          => 'Activate this checkbox when items that begin during the current settlement run shall be charged proportionately. E.g. if an monthly paid item starts in the middle of the month the customer would be charged only half of the full price in this settlement run.',
         'Type'                  => 'All fields besides Billing Cycle have to be cleared before a type change! Otherwise products can not be saved in most cases',
+        'deprecated'            => 'Active this checkbox if this product should not be showed as a valid product while creating/editing items.',
```
NinoRy:
```suggestion
        'deprecated'            => 'Activate this checkbox if this product shall not be shown in the product select list when creating/editing items.',
```

modules/Ticketsystem/Entities/Ticket.php
```diff
@@ -306,6 +323,31 @@ public function validGlobalSettings()
 
         return ['noReplyName' => $config->noReplyName, 'noReplyMail' => $config->noReplyMail];
     }
+
+    private function provideContractInformation()
```
cschra: this is Blade (view) code and does not belong here :innocent: :stuck_out_tongue_winking_eye:

modules/Ticketsystem/Entities/Ticket.php
```diff
@@ -165,6 +167,21 @@ public function contract()
         return $this->belongsTo(\Modules\ProvBase\Entities\Contract::class);
     }
 
+    public function contact()
```
cschra: this might be the perfect use case for a [Polymorphic One to Many](https://hackernoon.com/eloquent-relationships-cheat-sheet-5155498c209) relationship. [Laravel](https://laravel.com/docs/5.5/eloquent-relationships#polymorphic-relations) has a good documentation on this.

modules/Ticketsystem/Entities/Ticket.php
NinoRy: moved in 2badd8e

modules/Ticketsystem/Entities/Ticket.php
NinoRy: Would currently be a big effort (especially implementing it in a generic way) with no real value for any customer. 
(Store function had to be extended, select values in create/edit blade should be fetched via ajax to not have million of select values - could be crushing the browser)

Created issue for that: https://github.com/nmsprime/nmsprime/issues/886

PR #838, opened 2020-01-16
Bugfix/small fixes ole

olebowle:


modules/ProvBase/Entities/Modem.php
```diff
@@ -1777,6 +1777,10 @@ public function created($modem)
         $hostname = ($modem->isPPP() ? 'ppp-' : 'cm-').$modem->id;
         $modem->hostname = $hostname;
 
+        // always set hostname, even if updating() fails (e.g php warning)
+        // this is needed for a consistent dhcpd config
+        Modem::where('id', $modem->id)->update(['hostname' => $hostname]);
```
cschra: why not just call `$modem->save()`

olebowle: Because this calls updating() which we don't want at this point.

PR #846, opened 2020-01-22
Small bugfixes for 2.5.x

cschra:
OpenLayers:
* Keep Files that are potentially used in the future for compatibility
* there are svg and png files that are not used yet but may break
  OpenLayers functionality in the future...

All Networks:
* show a HTTP 422 Error page and display a custom error message


modules/HfcBase/Http/Controllers/TreeErdController.php
```diff
@@ -66,11 +66,22 @@ public function __construct()
...
+
         // prepare search query
-        $s = $field == 'all' ? 'id>2' : "$field='$search'";
+        if ($field == 'all') {
+            $field = 'id';
+            $operator = '>';
+            $search = 2;
+
+            if (! NetElement::where($field, $operator, $search)->count()) {
+                return redirect()->back();
```
NinoRy: Isn't it better to show a hint that the user has to add a net instead of quietly returning letting the user in the dark why nothing happens?

cschra: done

PR #847, opened 2020-01-22
Bugfix/small fixes par

xee8ai:
#


NinoRy:
What is the problem of the 'use Exception' statement?

PR #848, opened 2020-01-23
ERD performance

cschra:
It annoyed me that the ERD pages, especially the "All Nets", took so long to load.

On inspection I got over 970 Queries on TreeErd-View and 1600+ Queries for Topography view for "All Nets" per page load on a system with 200 Netelements.

This PR fixes this and reduces the queries to 6 in ERD view and 8 in Topography and therefore also the time for a page load significantly.

To get Stable this is possibly a Feature for 2.6.x?

### DIFF
--- a/modules/HfcBase/Http/Controllers/TreeErdController.php
+++ b/modules/HfcBase/Http/Controllers/TreeErdController.php
@@ -275,7 +275,7 @@ class TreeErdController extends HfcBaseController
                     $cri = $netelem->modems_critical_count;
                     $avg = $netelem->modemsUsPwrAvg;
 
-                    $file .= "\n node [label = \"$numa\\n$num/$cri\\n$avg\", shape = circle, style = filled, color=$color, URL=\"$url\", target=\"".$this->html_target.'"];';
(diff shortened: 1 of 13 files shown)


cschra:
@olebowle it might make sense to eager load icingaObjects as well to avoid another n+1 issue (1 DB query for each NetElement)

modules/HfcBase/Http/Controllers/TreeErdController.php
```diff
@@ -297,17 +296,17 @@ public function graph_generate($netelements)
         // TODO: Customer
         //
         if (\Module::collections()->has('HfcCustomer')) {
-            $n = 0;
+            $ModemHelper = \Modules\HfcCustomer\Entities\ModemHelper::class;
```
NinoRy: variable name should be camelcase or not?

cschra: This is a class name and these are typical PascalCase (also called StudlyCase).

Due to our modular architecture we can't "use" this class at the top, so to make it clear it is a class I thought it would be good to violate the standard "variable naming scheme".

If you don't agree I surely can refactor this towards camelCase.

PR #858, opened 2020-01-30
Performance improvements

cschra:
As the original PR #848 was not only about ERD performance any more, I changed the branch name.

This PR fixes mainly ERD - related issues, as this took forever to load and made a huge amount of unnecessary DB requests. But it also removes unused code in the Base Model and other Models to speed up NMS Prime by roughly factor 2.

There are still things to do:
- [x] Netgw and Modem index Page have 100+ Queries for Datatables -> SLA is checked for each model and not for the system - I recommend caching the result for the session? One SLA check uses 10 queries -> i.e. use withCount
- [ ] Tree View is also very query intensive, especially configfile and netelementtype
- [ ] MibFile has 24+ queries just counting oids -> use withCount...
- [ ] Global Config Page has some duplicate queries....

For Discussion: 
Fix the above mentioned speed/performance issues before merging or put them inside a ticket for the (unforeseeable) future?

In my Opinion, at least the SLA issue should be addressed.

### DIFF
--- a/app/BaseModel.php
+++ b/app/BaseModel.php
@@ -26,8 +26,6 @@ class BaseModel extends Eloquent
     // flag showing if children also shall be deleted on Model::delete()
     protected $delete_children = true;
 
-    public $external_voip_enabled;
-    public $billing_enabled;
     protected $fillable = [];
 
     public $observer_enabled = true;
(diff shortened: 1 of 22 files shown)


NinoRy:
maybe we should enable caching of the module config as well? Then e.g. Module::collections() will be cached. The config file is config/modules.php --> search for 'cache' key

cschra:
> maybe we should enable caching of the module config as well? Then e.g. Module::collections() will be cached. The config file is config/modules.php --> search for 'cache' key

we can certainly do that. But as far as the documentation states only the module.json files are cached. So Module::collections() will become a bit faster, but is still called every time. So to really fix these calls, the result of Module::collections() needs to be cached.

With this PR the mass use is avoided, for all the remaining checks I would instantiate a singleton object for the module states and this is then checked throughout the whole request lifecycle -> this would reduce the execution to only one per Request. Or we go for long term caching and when triggering the artisan commands to enable/disable, we bust the cache (which would make sense anyway)

Furthermore, we are still behind for Laravel Modules package (we are at v2.x and the most recent is v6.x). 

In my opinion a large part is done already and I expect only small speed wins. My recommendation would be to wait until we upgrade to Laravel 6 for the large changes, but I can make a commit to just enable module.json caching.

Issue #454, opened 2017-11-01
New user friendly Contract Page


Gitrosa:
<p>It should be easier and faster to add Modems, Items and SepaMandates on Contract page. It should be possible by 2 clicks (create - save) and filling only the most important data that is not filled by default.<br/>
(Nice to have: a warning could appear if the user leaves the page and there's no sepamandate assigned"</p>

<p>2 click mechanism:</p>
<ul>
	<li>reference system is that of albismart</li>
	<li>you click on "create model" and it opens the fields dynamically, that are important to work. This can also be a pop up window. Below the important fields a save button must appear</li>
</ul>

<p>important data of the models:</p>
<ul>
	<li>Modem: MAC, Configfile</li>
	<li>Item: Product</li>
	<li>SepaMandate: IBAN</li>
</ul>

<p>Note: Everything else should be filled with default values as it already is. If the user wants to change these he still can/must enter the appropriate model and change the default values by himself</p>
<i>[LAR-152] created by @NinoRy</i>

PR #860, opened 2020-02-01
Feature/base mvc 3 using vuejs

kokuta:
Issue: #454


NinoRy:
Hi. Just a reminder if you leave the testing stage ;) -> the current generic BaseController functions like edit() should be left as they are and new function names and routes need to be added so that we can have the old jquery concept in parallel to the new vue.js way. So we can switch all models step by step to the new version.

kokuta:
Hi, now creating ticket comments and retrieving them works through vue js.
I made a API Version 1 proposal since the Version 0 wasn't useful on the way I planned. I used two github resources and customized for our platform.
Please check https://github.com/esbenp/bruno, and https://github.com/esbenp/architect resources. I think they are a cool way to retrieve resources that frontend needs. 
Version1  uses the basic auth also as the v0. IMHO we should change to OAuth2(Passport).

kokuta:
This is replaced with pull request #917 .

PR #868, opened 2020-02-13
Datatable and Edit-Page fixes and Performance improvements

cschra:
#

### DIFF
--- a/app/Http/Controllers/GuiLogController.php
+++ b/app/Http/Controllers/GuiLogController.php
@@ -111,7 +111,7 @@ class GuiLogController extends BaseController
             })
             ->editColumn('created_at', function ($guilog) use ($routeName) {
                 return '<a href="'.route($routeName.'.edit', $guilog->id).
-                        '" title="'.str_replace(', ', '1&#013;', $guilog->text).'"><strong>'.
+                        '" title="'.e(str_replace(', ', '1&#013;', $guilog->text)).'"><strong>'.
                         $guilog->view_icon().$guilog->created_at.'</strong></a>';
             })
             ->rawColumns(['created_at'])
--- a/app/Http/Controllers/GuiLogController.php
+++ b/app/Http/Controllers/GuiLogController.php
@@ -102,15 +102,12 @@ class GuiLogController extends BaseController
         return DataTables::make($request_query)
             ->addColumn('responsive', '')
             ->setRowClass(function ($guilog) {
-                $bsclass = 'info';
-                if ($guilog->method == 'created') {
-                    $bsclass = 'success';
-                }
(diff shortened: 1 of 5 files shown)


modules/ProvBase/Entities/Modem.php
```diff
@@ -318,7 +318,7 @@ public function view_has_many()
         // we use a dummy here as this will be overwritten by ModemController::editTabs()
         if (Module::collections()->has('ProvVoip')) {
             $ret['Edit']['Mta']['class'] = 'Mta';
-            $ret['Edit']['Mta']['relation'] = $this->mtas;
+            $ret['Edit']['Mta']['relation'] = $this->mtas()->with('configfile')->get();
```
NinoRy: Take care! Please set relation for $this here as on next $this->mtas another DB query would be executed!

modules/ProvVoip/Entities/Mta.php
```diff
@@ -81,7 +81,7 @@ public function view_belongs_to()
     public function view_has_many()
     {
         $ret['Edit']['Phonenumber']['class'] = 'Phonenumber';
-        $ret['Edit']['Phonenumber']['relation'] = $this->phonenumbers;
+        $ret['Edit']['Phonenumber']['relation'] = $this->phonenumbers()->with('phonenumbermanagement')->get();
```
NinoRy: set relation here as well!

app/Http/Controllers/GuiLogController.php
```diff
@@ -111,7 +111,7 @@ public function filter($id, Request $request)
             })
             ->editColumn('created_at', function ($guilog) use ($routeName) {
                 return '<a href="'.route($routeName.'.edit', $guilog->id).
-                        '" title="'.str_replace(', ', '1&#013;', $guilog->text).'"><strong>'.
+                        '" title="'.e(str_replace(', ', '1&#013;', $guilog->text)).'"><strong>'.
```
NinoRy: Just to inform me. Since when does this function exist?

cschra: [Laravel 5.1 ](https://laravel.com/docs/5.1/helpers#method-e)

modules/ProvBase/Entities/Modem.php
cschra: Neat trick, in this particular case it wouldnt make any difference, because there is no other call, but I changed my implementation, as this could be beneficial in the future.

modules/ProvVoip/Entities/Mta.php
cschra: changed 👍

PR #869, opened 2020-02-16
bugfix/small-fixes-ole

olebowle:

### DIFF
--- a/app/Http/Controllers/Auth/LoginController.php
+++ b/app/Http/Controllers/Auth/LoginController.php
@@ -101,7 +101,7 @@ class LoginController extends Controller
 
         self::setDashboardNotifications();
 
-        $user->update(['last_login_at' => Carbon::now()]);
+        App\User::where('id', $user->id)->update(['last_login_at' => Carbon::now()]);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


modules/ProvBase/Entities/Modem.php
```diff
@@ -264,11 +264,6 @@ public function tree()
         return $this->belongsTo(\Modules\HfcReq\Entities\NetElement::class);
     }
 
-    public function nelelement()
```
cschra: > ```php   
> // TODO: deprecated! use netelement function instead - search for all places where this function is used
> public function tree()
> {
>     return $this->belongsTo(\Modules\HfcReq\Entities\NetElement::class);
> }
> ```
As by this comment it is advised to switch to the deleted function - either delete the comment or switch to the netelement function as recommended :wink:

NinoRy: looks like tree function is also not used. Can be deleted as well.

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -101,7 +101,7 @@ protected function authenticated(Request $request, $user)
 
         self::setDashboardNotifications();
 
-        $user->update(['last_login_at' => Carbon::now()]);
+        App\User::where('id', $user->id)->update(['last_login_at' => Carbon::now()]);
```
olebowle: @cschra Should we rather user ->save()? Did you want to circumvent the observer?

olebowle: We decided to circumvent the observer.

PR #875, opened 2020-03-10
show picture in modem analyses

royschn:


modules/ProvMon/Http/Controllers/ProvMonController.php
```diff
@@ -213,6 +214,26 @@ public function modemOnlineStatus($modem, $hostname)
...
+    /**
+     * Find matching picture of modem.model.
+     *
+     * @author  Roy Schneider
+     * @param Modules\ProvBase\Entities\Modem
+     * @return  string
+     */
+    private function modemPicture($modem)
+    {
+        foreach (\File::allFiles(public_path('images/modems')) as $file) {
```
olebowle: As far I can see the filenames are not sorted by default. Thus I would suggest using:
`collect(\File::allFiles(public_path('images')))->sortBy(function ($file) { return $file->getFilename(); });`

modules/ProvMon/Http/Controllers/ProvMonController.php
```diff
@@ -213,6 +214,26 @@ public function modemOnlineStatus($modem, $hostname)
...
+            preg_match('/(?<=\d-)(.*)(?=\.\w+)/', $file->getFilename(), $filename);
```
olebowle: I guess this is easier: `/\d+-(.+)\..+/`

modules/ProvMon/Http/Controllers/ProvMonController.php
olebowle: `images/modems` should be a const

PR #896, opened 2020-04-22
bugfix/small-fixes-nino

NinoRy:
#


olebowle:
What is our current convention regarding 'use'? Didn't we say, that if we use an include more than once we add the 'use' statement?

NinoRy:
i understood it that way if the class name is used more than once we definitely add the use statement otherwise it's open to the developer. But i can easily change it ;)

olebowle:
I don't mind.

PR #908, opened 2020-05-06
Fix bug in creating cdr LaTeX containing “1&1”.

xee8ai:
#


xee8ai:
Also prevent crashes on float values written in German syntax (comma instead of dot).

PR #917, opened 2020-06-23
Feature/api v1 setup

kokuta:
#


app/Http/Controllers/BaseController.php
```diff
@@ -239,26 +241,6 @@ protected function prepare_input($data)
...
     }
 
-    /**
-     * Normalizes numeric values to minimize problems for e.g. German users (using a comma instead a dot in float).
-     *
-     * @param $value the numeric string to normalize
-     *
-     * @author Patrick Reichel
-     */
-    protected function normalizeNumericString($value)
```
NinoRy, line 249: why did you remove this function when it it still used in Code?

kokuta: Removed by mistake during resolving conflicts.

PR #921, opened 2020-07-06
Feature/trouble dashboard

cschra:
#


modules/Dashboard/Resources/views/index.blade.php
```diff
@@ -81,15 +81,24 @@
 </div>
 <div class="row">
     <div class="col-xl-8 no-gutters ui-sortable">
-        @if($impairedData)
-            @section ('trouble_dashboard')
+        @if(! empty($impairedData))
```
NinoRy, line 68: empty space between @if and (

modules/Dashboard/Resources/views/index.blade.php
```diff
@@ -81,15 +81,24 @@
...
+            @section ('impaired_summary')
+                @include('HfcBase::troubledashboard.summary')
```
NinoRy: same here (empty space)

modules/HfcBase/Entities/IcingaServiceStatus.php
```diff
@@ -29,13 +29,6 @@ class IcingaServiceStatus extends Model implements ImpairedContract
      */
     protected $primaryKey = 'servicestatus_id';
 
-    /**
-     * Contains more detailed for Subservices, deserialized from perfdata field.
-     *
-     * @var \Illuminate\Support\Collection
-     */
-    public $additionalData;
```
NinoRy, line 37: why did you remove this when variable is still in use?

modules/HfcReq/Entities/NetElement.php
```diff
@@ -184,6 +184,7 @@ public function view_belongs_to()
     public function scopeWithActiveModems($query, $field = 'id', $operator = '>', $id = 2)
     {
         return $query->where($field, $operator, $id)
+            ->select(['id', 'id_name', 'name', 'ip', 'cluster', 'net', 'netelementtype_id', 'netgw_id', 'parent_id', 'link', 'descr', 'pos'])
```
NinoRy, line 187: why do you add this filter? Are you sure that there's nothing else accessed in all Contollers? This shouldn't be a perceptible performance improvement

cschra: 1. I checked the columns and there are currently all properties that are in use
2. The performance gain is in this case Not on the DB side, but the json that gets transfered is smaller (on live system it is around 250kb savings (1,45mb to 1,2mb)

modules/HfcBase/Entities/IcingaServiceStatus.php
cschra: I made it a dynamic property instead of a public property -> this ensures the property is in the encoded json string...

modules/HfcCustomer/Console/ClustersCommand.php
```diff
@@ -36,74 +89,53 @@ class ClustersCommand extends Command
...
     {
-        $status = 'OK';
-        $output = '';
-        $lookup = [
-            'OK' => 0,
-            'WARNING' => 1,
-            'CRITICAL' => 2,
-        ];
-
         foreach (NetElement::withActiveModems()->get() as $netelement) {
```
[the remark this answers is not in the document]
cschra: In a lot of cases where I use this scope I need all Netelements. (i.e. to load parents directly without another db call) lets see how it scales :-)

modules/HfcBase/Entities/IcingaObject.php
```diff
@@ -46,11 +46,29 @@ protected static function booted()
      *
      * @return \Illuminate\Database\Eloquent\Relations\HasOne
      */
-    public function hoststatus()
+    public function hostStatus()
```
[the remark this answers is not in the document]
cschra: done

modules/HfcReq/Entities/NetElement.php
cschra: introduced a parameter to use select only when needed and to increase clarity -> see 4a4a1f7

modules/Dashboard/Resources/views/index.blade.php
cschra: solved whitespace issues with d5af8c2

PR #922, opened 2020-07-15
Feature/dragdrop

Gitrosa:

### DIFF
--- a/modules/ProvBase/Resources/views/Configfile/edit.blade.php
+++ b/modules/ProvBase/Resources/views/Configfile/edit.blade.php
@@ -43,8 +43,8 @@
                                 <span class="dragdropitemmenubutton" v-on:click="moveItem(key, -1, id)">{{ trans('view.Button_DragDrop MoveToNewList') }}</span>
                                 <span class="dragdropitemmenubutton" v-on:click="moveItem(key, 0, id)">{{ trans('view.Button_DragDrop DeleteElement') }}</span>
                             </div>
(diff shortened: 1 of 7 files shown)


modules/ProvBase/Resources/views/Configfile/edit.blade.php
```diff
@@ -106,6 +151,20 @@
             //delete the list
             this.lists.splice(key, 1);
         },
+        ddFilter: function(key) {
+            var search=document.getElementById("ddsearch").value;
+            if (search != "") {
```
NinoRy, line 156: Just a note: We like to use
```suggestion
            if (search == "") {
                return null;
            }
```
here as it reduces tabs (intendation) in the following lines

modules/ProvBase/Resources/views/Configfile/edit.blade.php
```diff
@@ -73,6 +94,30 @@
...
+            targetElement=element.parentNode.getElementsByClassName("dragdropitemmenubox")[0];
+            if (targetElement.style.display!="block"){
+                targetElement.style.display="block";
+            }
+            else {
+                targetElement.style.display="none";
+            }
+        },
+        moveItem: function(olist, key, id) {
+            //for creating a new list
```
NinoRy, line 107: Just a note relating the format: We always keep a space between // and the comment
```suggestion
            // for creating a new list
```

NinoRy:
@Gitrosa please have a look at the comments and decide if you want to adapt sth to the recommendations or just leave it as is

PR #927, opened 2020-07-25
Feature/network graph

kokuta:
#


modules/HfcBase/Http/Controllers/VicinityGraphController.php
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Modules\HfcBase\Http\Controllers;
+
+use View;
+use Module;
+use App\Http\Controllers\BaseController;
+
+class VicinityGraphController extends BaseController
+{
+    public function showGraph(){
```
NinoRy: I would prefer the short name 'show' for this function (unless you have another function that is already called like this) as the controller is already called VicinityGraph and it's therefor a bit redundant and could be shorter ... also in Routes and so on.

config/hfcbase.php
```diff
@@ -1,7 +1,15 @@
...
 
-    return 		[
-        'name' => 'HfcBase',
-        'link' => 'HfcBase.index',
-        'MenuItems' => [],
-    ];
+use Modules\HfcReq\Entities\NetElement;
+
+return [
+    'name' => 'HfcBase',
```
NinoRy, line 6: The link to the vicinity graph shall appear as a "tab" of a customer bubble e.g. https://demo.nmsprime.com:8080/admin/Customer/netelement_id/33 (so no global menu entry)

modules/HfcBase/Resources/views/Tree/topo.blade.php
```diff
@@ -8,7 +8,6 @@
 	<script async defer src="{{asset('/modules/hfcbase/OpenLayers-2.13.1/OpenLayers.js')}}"></script>
 	<script async defer src="https://maps.google.com/maps/api/js?v=3.2&sensor=false&key={{config('app.googleApiKey')}}"></script>
 
-	@include ('HfcBase::Tree.topo-api')
```
NinoRy, line 11: why did you remove this line?

PR #937, opened 2020-08-10
Feature/user geopos

NinoRy:
#


routes/web.php
```diff
@@ -118,4 +118,9 @@
         'as' => 'Sla.clicked_sla',
         'uses' => 'SlaController@clicked_sla',
     ]);
+
+    BaseRoute::post('user/updateGeopos', [
+        'as' => 'user.updateGeopos',
+        'uses' => 'Auth\UserController@updateGeopos',
+    ]);
```
olebowle, line 125: it seems the middleware is missing

schmto: does BaseRoute include middleware?

PR #946, opened 2020-08-11
feature/user-geopos

NinoRy:
#


cschra (approved):
Are there other positions than the geoposition? In my Opinion just "last position update" would be sufficient...
but codewise there are no issues, hence I am approving this PR.

PR #949, opened 2020-08-25
bugfix/small-fixes-nino

NinoRy:


NinoRy:
Note for e2341d3
ticket_type table is our first and only special case where we follow laravel convention where table name is split by underscore because of camelcase notation of the model ('TicketType') - our basemodel functionality didn't respect this so far. We should do this when there are more tables following, but as of now i just added a special treatment.

royschn (approved):

